PR #186, opened 2015-04-14
Allow user to override the value of $virtualenv in python::virtualenv

bakks:
I'm having an issue with `python::virtualenv`: when I set `$version` to `pypy` it attempts to create the virtual enviroment with `virtualenv-pypy`. I'm not sure where that binary is meant to come from, but at least on the platform I'm using (CentOS 7) pypy doesn't ship with a `virtualenv-pypy` command. So the point of this pull req is to allow the user to choose which `virtualenv` command to use. Happy to make any necessary modifications, please take a look.


shivapoudel:
cool!

PR #187, opened 2015-04-15
use mode in dotfile file resource

jlyheden:


igalic:
how'd that ever work??

PR #191, opened 2015-04-17
Fix duplicate virtualenv declaration

yrobla:
Tests on this module are failing with

Error: Cannot reassign variable virtualenv at
/etc/puppet/modules/python/manifests/virtualenv.pp:102

According to puppet, variable reassignment is not allowed:
https://docs.puppetlabs.com/puppet/latest/reference/
lang_variables.html#no-reassignment

Using a new var for the reassingment to avoid that problem

### DIFF
--- a/manifests/virtualenv.pp
+++ b/manifests/virtualenv.pp
@@ -99,10 +99,13 @@ define python::virtualenv (
     }
 
     if $virtualenv == undef {
-      $virtualenv = $version ? {
+      $used_virtualenv = $version ? {
         'system' => 'virtualenv',
         default  => "virtualenv-${version}",
       }
+    }
+    else {
+      $used_virtualenv = $virtualenv;
     }
 
     $proxy_flag = $proxy ? {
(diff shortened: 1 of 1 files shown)


shivapoudel:
@yrobla Do you need this for what?

yrobla:
We are using this module on review.openstack.org.
We tried to bump to the latest version of the module , 1.9.2. My colleague @nibalizer sent a fix for that to correct paths yesterday and we wanted to use that. But when we applied that version into the tests, it's failing due to that reassignment of var. That's why i was trying to get that fixed.
Thanks

shivapoudel:
cool!

madphenomenon:
I have the same problem:
> Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Cannot reassign variable virtualenv at /etc/puppet/environments/testing/modules/python/manifests/virtualenv.pp:102 on node test

madphenomenon:
@shivapoudel after upding https://forge.puppetlabs.com/stankevich/python to a 1.9.3 version I see the error:
> Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Syntax error at ';'; expected '}' at /etc/puppet/environments/testing/modules/python/manifests/virtualenv.pp:108 on node test

code in this line:
> 107    else {
> 108     $used_virtualenv = $virtualenv;
> 109   }
must be
> 107    else {
> 108     $used_virtualenv = $virtualenv,
> 109   }

shivapoudel:
@madphenomenon this is shipped in `v1.9.4`. Please updated :)

madphenomenon:
@shivapoudel thank you! :)

PR #192, opened 2015-04-17
Fix typo in virtualenv

yrobla:
Ending ; was causing tests to not pass, amending that.


yrobla:
Tests were not passing, i did a typo. I didn't expect that to be merged so fast, thx for it and sorry for the bug.

Issue #115, opened 2014-08-14
include epel for RedHat


lesaux:
Should we include epel when running on Rehat family?

shivapoudel:
Closed in favor of #198

PR #198, opened 2015-06-13
Add EPEL dependency

### LINKED ISSUE #196: python-pip requires EPEL on Redhat/CentOs 6 and 7
First, apologies for opening this issue. There's an existing issue #115 which begins to describe this bug. I'm opening this issue because conversation on #115 has been restricted to contributors by @shivapoudel (which is of course totally fine).

@shivapoudel would it be possible to add a comment to that issue to indicate the status of any efforts to fix that bug or what people should do who use RHEL/CentOS? From the comments in that issue I wasn't able to determine why the conversation had been restricted (there may be background to this issue I'm unaware of).

To clarify the problem (which isn't very specific in #115), under RHEL/CentOS 7, attempting to install python-pip with a stanza like this 

```
  class { 'python' :
    pip => true,
  }
```

will fail because the [system package that this puppet module calls for](https://github.com/stankevich/puppet-python/blob/09967704c152f5fd044ed4ddbc7382ef2e24aea5/manifests/install.pp#L60) isn't part of the base RHEL/CentOS packages, but instead part of [EPEL](https://fedoraproject.org/wiki/EPEL).

A solution to this problem (which I'll try to PR) would be to include the [stahnma-epel](https://forge.puppetlabs.com/stahnma/epel) puppet module as a dependency in the [metadata.json](https://github.com/stankevich/puppet-python/blob/09967704c152f5fd044ed4ddbc7382ef2e24aea5/metadata.json#L43-L48). The nice thing about the `stahnma-epel` puppet module is that it only takes effect on [RHEL derivatives](https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux_derivatives) and [does nothing on other operating systems](https://github.com/stahnma/puppet-module-epel/blob/7322b80c268087bc4b967a9a6ca29923853971f8/manifests/init.pp#L132-L134).

gene1wood:
This approach to adding the EPEL dependency uses a few more lines of code than the alternative, but it means that the EPEL module only gets called if the OS (RedHat) and OS Version require it.

Note: Even if the EPEL module were included on a different OS (e.g. Ubuntu) the module logic of the EPEL module doesn't do anything unless it's on a RedHat osfamily system.

An alternative method to accomplish this would be to just unconditionally include the EPEL module, and set it as a require for the pip and virtualenv package resources. On Ubuntu this would mean that puppet output would show the EPEL module being used but nothing would be done to the system. On RedHat the EPEL yum repo configs would be installed even if the user was on RHEL 7 and only looking to install virtualenv (not pip) in which case EPEL is not required because RHEL 7 provides virtualenv in the base repositories.

I figured that the method in this PR would be better than the alternative as it would only involve the EPEL module when required, even though that resulted in a couple more lines of conditional logic.

One other thing to note is that regardless of the OS (Ubuntu RedHat), the EPEL module would be downloaded because this PR sets it as a metadata.json dependency. This means on an Ubuntu system, puppet (or librarian-puppet or whatever you use to fetch puppet modules) would download the EPEL module, but would not use it. It would just be in the modules directory unused. I can't think of a way around this because module dependencies (like those recorded in metadata.json) aren't conditional and can't depend on the OS. I don't think it's a big problem though.

Fixes #196

### DIFF
--- a/Modulefile
+++ b/Modulefile
@@ -9,3 +9,4 @@ project_page  'https://github.com/stankevich/puppet-python'
 
 ## Add dependencies, if any:
 dependency    'puppetlabs/stdlib', '>= 4.0.0'
+dependency    'stahnma/epel', '>= 1.0.1'
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -56,6 +56,16 @@ class python::install {
       package { "python==${python::version}": ensure => latest, provider => pip }
     }
     default: {
+      if $::osfamily == 'RedHat' {
+        if $pip_ensure == present {
+          include 'epel'
+          Class['epel'] -> Package[$pip]
+        }
+        if ($venv_ensure == present) and ($::operatingsystemrelease =~ /^6/) {
+          include 'epel'
+          Class['epel'] -> Package['python-virtualenv']
+        }
+      }
       package { 'python-virtualenv': ensure => $venv_ensure }
       package { $pip: ensure => $pip_ensure }
       package { $pythondev: ensure => $dev_ensure }
--- a/metadata.json
+++ b/metadata.json
@@ -44,6 +44,10 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">= 4.0.0"
+    },
+    {
+      "name": "stahnma/epel",
+      "version_requirement": ">= 1.0.1"
     }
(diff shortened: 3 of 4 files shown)


gene1wood:
This PR may also address #115

PR #202, opened 2015-06-16
Confine puppet facter to Linux

gajadevops:
This is to make sure it runs on linux os only. We do have stacks comprised of windows & linux machines

### DIFF
--- a/lib/facter/pip_version.rb
+++ b/lib/facter/pip_version.rb
@@ -15,6 +15,7 @@ else
 end
 
 Facter.add("pip_version") do
+  confine :kernel => :linux
   has_weight 100
   setcode do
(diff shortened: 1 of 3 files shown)


shivapoudel:
Nop, I do not prefer this.

gajadevops:
Hi shivapoudel,
Whats the reason?

gajadevops:
Do you think this is the better way ?

if Facter.value(:kernel) != "windows"
 Facter.add("pip_version") do
   has_weight 100
   setcode do
     if Facter::Util::Resolution.which('pip')
       Facter::Util::Resolution.exec('pip --version 2>&1').match(/^pip (\d+\.\d+\.?\d*).*$/)[1]
     end
   end
 end
end

PR #203, opened 2015-06-19
Add support for SLES

buzzdeee:
Besides added Support, add specs for SLES, and update metadata.json
to reflect the support.

Developed and tested on SLES11 SP3. 
Let me know if there is something to change, enhance, etc. and I'm happy to do so.


shivapoudel:
@buzzdeee Great work!

PR #208, opened 2015-07-04
Deprecate modulefile

ghoneycutt:


ghoneycutt:
Suggest just merging #209 which includes these commits

shivapoudel:
Closed in favour of #209

PR #210, opened 2015-07-05
remove broken facts

ghoneycutt:


ghoneycutt:
Agree that the version must be determined from the binary itself as opposed to determining from a package.

Issue #189, opened 2015-04-16
Support for SCL?


erik-stephens:
I don't see a way to use this module with a python from [SCL](https://www.softwarecollections.org).  Anyone know how, if at all, to get this to work with an SCL python?  If not, is it something you would like to support in this module?  I can work on a PR if would be useful.

Tokynet:
BUMP :)

shivapoudel:
@erik-stephens PR welcome :)

sorreltree:
A few general notes on this that didn't go into the pull request.

The pyvenv module appears to insist on creating the target directory for the venv before running the pyvenv command.  This actually breaks pyvenv unless you pass in the --clear or the --update command in python 3.3.  I kludged this by putting --clear in the pyvenv command for scl, but I think it will break for non-scl installs as well.  I'd say a better solution would be to either remove target directory resource or to include --clear in the default command line.

Additionally, there doesn't seem to be any order enforced on the creation of the venv, so you can end up with puppet trying to create the venv before the desired python version is installed.  This is of course fixable with the "play it again, Sam" method (just run puppet again and it'll do it) but that isn't the nicest of solutions.  I can open these as separate issues, but I wanted to note them as issues with the build here.

Finally, because of the weird way that RH designed SCL, you still need to either run scl enable or set the environment variables by hand before activating the virtualenv, but that's a "feature" of SCL.  I've considered adding an option to modify the activate file to do that automatically (because why not?) but that's a bigger undertaking.

PR #214, opened 2015-07-21
Experimental support for SCL repositories on RHEL and CentOS

sorreltree:
Resubmitting, hopefully passing Travis CI tests this time.  Sorry for the noise.

This is in response to issue #189 and provides nominal support for SCL packages. I have only had a chance to test this with CentOS 6 and the python33 collection, so please treat this as experimental and intended to be a conversation starter, not a completely finished patch.

The module appears to function nominally correctly by setting python::provider to 'scl' and python::version to 'python33' (or whatever the name of the collection is).

I'm not sure pip is doing the correct thing -- RedHat in its wisdom declined to include a pip package in the RHEL 6 SCLs

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,14 @@ python::python_pips:
     virtualenv: "/opt/env2"
 ```
 
+### Using SCL packages from RedHat or CentOS
+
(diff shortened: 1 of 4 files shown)


sorreltree:
Groovy.

PR #217, opened 2015-07-21
Update docs to reflect current pip parameter default and move release notes to the bottom of the page

### LINKED ISSUE #160: Documentation conflicts itself on whether or not pip must be explictly specified.
Both of these are specified:

"You must explicitly specify pip => true in the python class if you want pip installed. As such, the pip package is now independent of the dev package and so one can exist without the other."

"pip - Install python-pip. Default: true"

gene1wood:
Since the release notes reflect changes from long in the past it's probably confusing to have them at the very top of the page. I've moved them to the bottom to reflect the fact that they are not as important as the installation and usage instructions.
Fixes #160


gene1wood:
If you'd like I can throw this PR away and create a new one with only a single commit. Let me know if that's preferred.

PR #222, opened 2015-08-01
Added exec_prefix to clean up exec statements, particularly for scl

sorreltree:
This is a bit of a code cleanup that benefits the SCL provider code but also centralizes the specifics of exec in several circumstances to simplify some exec statement logic. No features added, just a tidy-up.

One change of note is that pyvenv now runs with the --clear flag.  In python 3.3, if the virtual environment directory already exists, pyvenv will fail.  In python 3.4, pyvenv will assume the --clear flag.  So this clears an error without changing working behavior.


shivapoudel:
@stankevich add the @petems as the collaborator :)

PR #229, opened 2015-08-21
Support Ruby versions 1.8.7 and 2.1.0

ghoneycutt:


shivapoudel:
@ghoneycutt anyway here we need the patch for Travis-CI

ghoneycutt:
After I get this right, adding support for the future parser and then looking at Puppet v4

ghoneycutt:
@shivapoudel good to merge :)

PR #238, opened 2015-09-07
Check that we have results before returning a value

xaque208:
Without this change, empty results yield a message like the following in
Facter output.

  undefined method `[]' for nil:NilClass

This work captures the results and check to them to ensure we return
a valid value.


shivapoudel:
Cool!

xaque208:
Thanks!

xaque208:
More to come if you're open to it.

shivapoudel:
xD

PR #239, opened 2015-09-08
Initial FreeBSD support

xaque208:
Here is the result of my effort to get Python managed on my FreeBSD systems.  Python 3 has some edge cases, and lacks packages for a few items on FreeBSD, so there is some logic to work around some of this.  I tried to keep it as clean as I could and still keep the main structure of the module.  See the commits for more information.

### DIFF
--- a/manifests/dotfile.pp
+++ b/manifests/dotfile.pp
@@ -39,7 +39,7 @@ define python::dotfile (
   $ensure   = 'present',
   $filename = $title,
   $owner    = 'root',
-  $group    = 'root',
+  $group    = '0',
   $mode     = '0644',
   $config   = {},
 ) {
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -25,18 +25,45 @@ class python::install {
     'RedHat' => "${python}-devel",
     'Debian' => "${python}-dev",
     'Suse'   => "${python}-devel",
+    default  => undef,
   }
 
-  $python_virtualenv = $::lsbdistcodename ? {
-    'jessie' => 'virtualenv',
-    default  => 'python-virtualenv',
+  case $::kernel {
+    'Linux': {
+      $python_virtualenv = $::lsbdistcodename ? {
+        'jessie' => 'virtualenv',
+        default  => 'python-virtualenv',
+      }
+    }
+    'FreeBSD': {
+      if $::python::version =~ /^3/ {
+        if ! $::python::pip {
+          notify { 'freebsd_virtualenv_needs_pip':
+            message => 'FreeBSD with Python 3 requires pip to install virtualenv; set pip => true',
+          }
+        }
+        $python_virtualenv = 'virtualenv'
+      } else {
+        $python_virtualenv = 'py27-virtualenv'
+      }
+    }
   }
 
   # pip version: use only for installation via os package manager!
-  if $::python::version =~ /^3/ {
-    $pip = 'python3-pip'
-  } else {
-    $pip = 'python-pip'
+  case $::kernel {
+    'Linux': {
+      if $::python::version =~ /^3/ {
+        $pip = 'python3-pip'
+      } else {
+        $pip = 'python-pip'
+      }
+    }
+    'FreeBSD': {
+      if $::python::version == 'system' or $::python::version =~ /^2/ {
+        $pip = 'py27-pip'
+      }
+    }
+    default: { }
   }
 
   $dev_ensure = $python::dev ? {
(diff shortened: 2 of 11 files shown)


xaque208:
I hope this is sane enough.  If any of this code needs adjustment for style or other reason to be accepted, please let me know and I will work to improve.  Some things are not ideal, but that is more a reflection on the state of Python on FreeBSD than anything else.  I'm happy to help improve this over time, and I've not tested everything, but I'm willing to maintain the FreeBSD portions of this code, as I'm expecting to use this across my fleet.

manifests/dotfile.pp
```diff
@@ -39,7 +39,7 @@
   $ensure   = 'present',
   $filename = $title,
   $owner    = 'root',
-  $group    = 'root',
+  $group    = '0',
```
ghoneycutt: Why is this needed for FreeBSD?

spec/classes/python_spec.rb
```diff
@@ -271,4 +271,101 @@
...
+      end
+    end
+
+    context "with python 3" do
+      let (:params) {{:version => '3'}}
+      it { is_expected.to contain_package('python3')}
+
+      context "with pip" do
+        let (:params) {{:version => '3', :pip => true}}
+        it { is_expected.to contain_exec('install_pip34')}
```
ghoneycutt: The resources in the spec tests should have their attributes listed along with their values, though given that the existing code is not doing this, I think it is OK.

ghoneycutt:
@xaque208 Thanks Zach!

The code looks great, though we should have someone else confirm that it is working for them. Anyone else have FreeBSD to test with?

ghoneycutt:
@xaque208 Could you also update the metadata.json in your last commit?

manifests/dotfile.pp
xaque208: There is no group 'root', basically.

xaque208:
The metadata has been updated.  Let me know if there is anything else that blocks acceptance of this change.

ghoneycutt:
@xaque208 Nothing else I can see that would block. We just need someone else to verify it works.

manifests/dotfile.pp
igalic: `wheel` :P

manifests/install.pp
```diff
@@ -142,9 +169,39 @@
...
+
+      if $python_virtualenv {
+        if $::osfamily == 'FreeBSD' and $::python::version =~ /^3/ {
+          package { $python_virtualenv:
+            ensure   => $venv_ensure,
+            provider => 'pip',
+          }
+        } else {
+          package { $python_virtualenv: ensure => $venv_ensure }
+        }
```
igalic: not a big fan of this if-pyramid.
can't we just add a variable `$provider` and override it?
that would make for more declarative looking code.

ghoneycutt: @igalic I'm happy to accept as is and then improve later. It should be tested on that actual platform though. Would you be able to test?

xaque208: @igalic I'm not much of a fan of that either.  I'd happily take a PR on the branch if you can help me work out a way to make that cleaner.  I'd rather avoid the `provider => $::kernel ? {...}` sort of junk.

ghoneycutt:
Hi @xaque208 

It seems that no one is stepping up for testing FreeBSD and I do not want to lose the work that you've done to support it as this code base is evolving quite quickly. If you could rebase this, I'll be sure to get it merged.

xaque208:
Well, thats a bummer.  It looks like the install class has been rewritten in the last two weeks.  My interest here is waning, but I'll take a look at a rebase.

xaque208:
I've looked at this.  My interest in updating this module was driven primarily based on the relatively simple approach this module took to python installation, relative to the complexity of the problem.  It seems that the installation has gotten more complicated in the last two weeks, and I can't quite tell the benefit.  A rewrite of this work is nearly required to accomplish what this PR originally intended to do, and I don't plan to spend the time and energy doing so.

I'll be keeping a fork to suit my needs for the time being, perhaps finding another module later.  Sorry for the noise.

https://github.com/xaque208/puppet-python/pull/1

ghoneycutt:
I'd have to agree with @xaque208 

In the environments that I usually work in, downloading things from the internet directly from a host for installation is unheard of. Instead we generally build packages with FPM and use the system's package manager. It would be better if this module took that approach and focused more on making python awesome than figuring out how to install things.

shivapoudel:
>It would be better if this module took that approach and focused more on making python awesome than figuring out how to install things.

@ghoneycutt If you could do it ;-)

PR #244, opened 2015-09-15
Bootstrap pip installation

### LINKED ISSUE #243: Unable to install pip using pip provider
Pull request #241 made it possible to install `pip` via `pip` with the following manifest:

```
class { 'python':
  ensure => 'present',
  provider  => 'pip',
}
```

This does not quite work as expected, however, because it presents a bootstrapping issue (it is not possible to install `pip` via `pip` if `pip` is not installed).

I'm not sure what the best way to fix this is...

joshuaspence:
Fixes #243.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -65,6 +65,14 @@ class python::install {
 
   case $python::provider {
     pip: {
+      # Install pip without pip, see https://pip.pypa.io/en/stable/installing/.
+      exec { 'bootstrap pip':
+        command => 'curl https://bootstrap.pypa.io/get-pip.py | python',
+        creates => '/usr/local/bin/pip',
+        require => Package['python'],
+      }
+      Exec['bootstrap pip'] -> Package <| provider == pip |>
+
       Package <| title == 'pip' |> {
         name     => 'pip',
         provider => 'pip',


manifests/install.pp
```diff
@@ -65,6 +65,14 @@
 
   case $python::provider {
     pip: {
+      # Install pip without pip, see https://pip.pypa.io/en/stable/installing/.
+      exec { 'bootstrap pip':
+        command => 'curl https://bootstrap.pypa.io/get-pip.py | python',
+        creates => '/usr/local/bin/pip',
```
joshuaspence: This is probably not platform independent.

shivapoudel: @joshuaspence ^^

joshuaspence: ?

shivapoudel: If you do it @joshuaspence :-)

manifests/install.pp
```diff
@@ -65,6 +65,14 @@
 
   case $python::provider {
     pip: {
+      # Install pip without pip, see https://pip.pypa.io/en/stable/installing/.
+      exec { 'bootstrap pip':
+        command => 'curl https://bootstrap.pypa.io/get-pip.py | python',
```
ghoneycutt: Downloading scripts from the internet and running them makes baby jesus cry. Can't we just install the packages?

manifests/install.pp
joshuaspence: Ok

manifests/install.pp
joshuaspence: This is only done if you choose `pip` as a provider.

Generally I'd agree with you, but I couldn't think of a better way to do this whilst remaining cross-platform compatible.

PR #245, opened 2015-09-15
Minor tidying

joshuaspence:
Various stylistic fixes.


shivapoudel:
Do you need this for what?

PR #252, opened 2015-09-22
Fix pip git repo install every puppet run

tremblaysimon:
I think it's a valid proposal to fix that bug since url seems to always be present when doing a pip freeze.

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -134,18 +134,14 @@ define python::pip (
     fail('python::pip cannot provide uninstall_args with ensure => present')
   }
 
-  # Check if searching by explicit version.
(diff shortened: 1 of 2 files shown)


tremblaysimon:
After thinking about that, there is still at least one problem with my fix: instead of installing always the version wanted, it will not install anything if the package is already installed. This behavior might not be wanted when someone want to update to latest/another version of the same package.

shivapoudel:
@tremblaysimon I am happy to merge if you could solve issue :-)

tremblaysimon:
The only thing I see to update the package for now is to handle commit/revision id. It supposes that pip install manifest verify git/svn/hg/bzr repository latest branch commit/revision id against what it is installed now (from pip freeze we get commit/revision id). Does it make sense?

PR #261, opened 2015-11-18
Bug virtualenv instead of virtualenv-$version

Asher256:
Hello,

Thank you for your useful Puppet module!

This pull request solves an error caused by the class "python::virtualenv", when the parameter "version" is changed (if you put something else than "system" in the parameter "version", "virtualenv-$version" is used instead of "virtualenv").

The error:
```
	Error: sh: 1: virtualenv-2: not found
	sh: 1: /opt/pyenv2/bin/pip: not found
	sh: 1: /opt/pyenv2/bin/pip: not found
	Error: /Stage[main]/Profiles::Python/Python::Virtualenv[pyenv2]/Exec[python_virtualenv_/opt/pyenv2]/returns: change from notrun to 0 failed: sh: 1: virtualenv-2: not found
	sh: 1: /opt/pyenv2/bin/pip: not found
	sh: 1: /opt/pyenv2/bin/pip: not found
```

My operating system: Ubuntu Wily

My Puppet code:
```
  python::virtualenv { 'pyenv2' :
    ensure       => present,
    version      => '2',
    venv_dir     => '/opt/pyenv2',
    cwd          => '/opt/pyenv2',
    timeout      => 0,
  }
```

Best regards, 

Asher256

### DIFF
--- a/manifests/virtualenv.pp
+++ b/manifests/virtualenv.pp
@@ -97,10 +97,7 @@ define python::virtualenv (
     }
 
     if $virtualenv == undef {
-      $used_virtualenv = $version ? {
-        'system' => 'virtualenv',
-        default  => "virtualenv-${version}",
-      }
+      $used_virtualenv = 'virtualenv'
     } else {
       $used_virtualenv = $virtualenv
     }


dansajner:
@shivapoudel @ghoneycutt I'm a bit unhappy with this change.  Obviously it was an issue for @Asher256 but the `virtualenv` parameter is available to specify the executable name.

The previous implementation was correct for my (any probably other's) use case where the executable name is indeed `virtualenv-$version`.  To accept this change all of our `python::virtualenv` declarations will need updated to pass along the `virtualenv` parameter including other upstream modules (namely https://github.com/evenup/evenup-beaver/blob/master/manifests/package.pp#L38-L44).

Would you consider reverting this patch in favor of not breaking the existing functionality for others?  Otherwise please consider this a breaking change in your next release.  Thanks!

Asher256:
An alternative solution dansajner would be to add an argument the users can use to specify the virtualenv executable name (or path).

Issue #262, opened 2015-11-20
Call, 'versioncmp' parameter 'a' expects a String value, got Undef


jonlil:
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, 'versioncmp' parameter 'a' expects a String value, got Undef at /etc/puppetlabs/code/environments/production/modules/python/manifests/virtualenv.pp:122:11

puppet 4.2, puppet-python 1.10.0

jonlil:
i'm using ubuntu 14.04

anthonymf:
I found this bug too.  I think the problem is this: the virtualenv_version custom facter fact fails to return a version string if virtualenv is not installed.

I haven't got time to raise a PR but I think this is a fix:

https://github.com/stankevich/puppet-python/blob/c2ca9d7495375bde781d766b3e6d3dd59cfbbed3/lib/facter/virtualenv_version.rb#L6
Change
```Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(/^(\d+\.\d+\.?\d*).*$/)[0]```
to
```Facter::Util::Resolution.exec('virtualenv --version 2>/dev/null || echo 0.0.0').match(/^(\d+\.\d+\.?\d*).*$/)[0]```

anthonymf:
:-)

PR #267, opened 2015-11-24
Revert "Update virtualenv_version.rb"

shivapoudel:
Reverts stankevich/puppet-python#265


rauchrob:
Why has this been reverted? #262 is still an issue for me.

shivapoudel:
If you could send the fix with travis-ci build pass, I am happy to merge :)

rauchrob:
Stupid question: shouldn't `python::virtualenv` fail intentionally if the `virtualenv_version` fact is not available?

PR #270, opened 2015-12-06
bugfix install pip on centos6 using scl

netors:


shivapoudel:
Fix the travis-ci builds ;)

netors:
done!

adamcstephens:
Why didn't you just use the SCL pip package?

PR #274, opened 2015-12-07
new manage_requirements argument to address issue 273

rosenbergj:


rosenbergj:
(My first ever pr; please be gentle...)

ghoneycutt:
@rosenbergj thanks for contributing!

manifests/requirements.pp
```diff
@@ -38,6 +38,9 @@
 # [*extra_pip_args*]
 # Extra arguments to pass to pip after the requirements file
 #
+# [*manage_requirements*]
```
ghoneycutt: This also needs to be documented in the README.md

manifests/requirements.pp
```diff
@@ -110,7 +114,7 @@
 
   # This will ensure multiple python::virtualenv definitions can share the
   # the same requirements file.
-  if !defined(File[$requirements]) {
+  if !defined(File[$requirements]) and $manage_requirements {
```
ghoneycutt: Add the `== true` bit makes it easier to understand exactly what is happening.

```puppet
if !defined(File[$requirements]) and $manage_requirements == true {
```

manifests/requirements.pp
rosenbergj: Fixed.

manifests/requirements.pp
rosenbergj: Good idea. Done.

ghoneycutt:
Looks good. We also need some spec tests - See https://github.com/stankevich/puppet-python/blob/master/spec/defines/requirements_spec.rb

Add a block where you let params set manage_requirements to false and it should test that the file `$requirements` is not in the catalog

``` ruby
context "/requirements.txt" do
        let (:params) {{ :requirements => "/requirements.txt", :manage_requirements => false }}
        it { is_expected.not_to contain_file("/requirements.txt") }
      end
```

rosenbergj:
Thank you!

PR #280, opened 2016-01-05
Force virtualenv_version to be a string.

dansajner:
This is a different approach to resolving https://github.com/stankevich/puppet-python/issues/262


ghoneycutt:
Thank you @dansajner !

PR #296, opened 2016-04-07
Name the pip package 'python-pip', as that is what others expect

mpdude:
For example, other modules check

```
  if !defined(Package['python-pip']) {
     package { 'python-pip': ... }
  }
```

This does not work if this module names the definition "pip" only.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -58,7 +58,7 @@ class python::install {
   case $python::provider {
     pip: {
 
-      package { 'pip':
+      package { 'python-pip':
         ensure  => $pip_ensure,
         require => Package['python'],
       }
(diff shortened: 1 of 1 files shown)


manifests/install.pp
```diff
@@ -58,7 +58,7 @@
   case $python::provider {
     pip: {
 
-      package { 'pip':
+      package { 'python-pip':
         ensure  => $pip_ensure,
```
ghoneycutt: would need to set the `name` attribute to be `'pip'` to retain current functionality.

manifests/install.pp
```diff
@@ -163,7 +163,7 @@
 
     default: {
 
-      package { 'pip':
+      package { 'python-pip':
```
ghoneycutt: same here

bastelfreak:
Hi @mpdude, this module got recently migrated to Vox Pupuli. Are you still interested in this PR (I know, it's been two years)? If so, could you please take a look at the comments @ghoneycutt made and rebase against our current master branch?

bastelfreak:
Hello. I'm going to close this PR due to inactivity. Please reopen and rebase it if you're still interested in it.

PR #297, opened 2016-04-11
EPEL only makes sense on RH systems but not Fedora

jcpunk:
This update to params.pp only sets EPEL true if it would make sense

### DIFF
--- a/spec/classes/python_spec.rb
+++ b/spec/classes/python_spec.rb
@@ -90,6 +90,13 @@ describe 'python', :type => :class do
         it { is_expected.to contain_package("python-dev").with_ensure('absent') }
       end
     end
+
+    describe "EPEL need not apply" do
+      context "default/empty" do
+        it { should_not contain_class('epel') }
+      end
+    end
+
   end
 
   context "on a Redhat 5 OS" do
@@ -111,6 +118,12 @@ describe 'python', :type => :class do
     it { is_expected.to contain_package("pip") }
     # Basic python packages (from pip)
     it { is_expected.to contain_package("virtualenv")}
+
+    describe "EPEL may be needed" do
+      context "default/empty" do
+        it { should contain_class('epel') }
+      end
+    end
 
     describe "with python::dev" do
       context "true" do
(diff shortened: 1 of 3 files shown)


ghoneycutt:
+1 Though there should be a spec test that tests for the inclusion of the epel class based on the platform.

I'll look into the issue with ruby 1.8.7 failing.

ghoneycutt:
working in PR #298 to fix ruby 1.8.7

ghoneycutt:
If you rebase off of master, it will pull in the changes necessary to get ruby 1.8.7 working for this PR.

jcpunk:
I have no idea how to test for the state of a variable withing a class.  Any pointers?

ghoneycutt:
You are not testing for a variable, which cannot be done, instead you are testing that a resource is or is not included in the catalog. So the tests would look like

```ruby
it { should contain_class('epel') }
```

and

```ruby
it { should_not contain_class('epel') }
```

spec/classes/python_spec.rb
```diff
@@ -90,6 +90,13 @@
         it { is_expected.to contain_package("python-dev").with_ensure('absent') }
       end
     end
+
+    describe "EPEL need not apply" do
+      context "default/empty" do
```
ghoneycutt: need to set the facts with values for osfamily and operatingsystem. Will need three different tests to correspond with the code in params.

```ruby
describe 'epel class' do
  context 'when osfamily is RedHat and operatingsystem is Fedora' do
    ...
  end
  context 'when osfamily is RedHat and operatingsystem is Fedora' do
    ...
  end
  context 'when osfamily is RedHat and operatingsystem is not Fedora' do
    ...
  end
  context 'when osfamily is not RedHat' do
    ...
  end
end
```

to set the facts, it would look like this

```ruby
context 'whatever' do
  let(:facts) do
    {
      :osfamily        => 'RedHat',
      :operatingsystem => 'Fedora',
    }
end
```

jcpunk:
In this last commit, I believe I've got the tests with the right facts set inserted along with their operating systems.

ghoneycutt:
Released in 1.18.1

Issue #234, opened 2015-08-22
SCL python27: add a workaround for libpython2.7.so.1.0 issue (LD_LIBRARY_PATH)


iakovgan:
Unsing SCL python27 we frequently run into an isssue with python when it can't find libpython2.7.so.1.0. Looks like there are a lot of ways of handeling this issue, but the simpliest was just adding an env variable LD_LIBRARY_PATH. I suppose it is a well known issue(see stackoverflow link in the code). 

Here is a workaround:  
```puppet
  # Set env variable LD_LIBRARY_PATH
  # see http://stackoverflow.com/questions/7880454/python-executable-not-finding-libpython-shared-library
  if $python::python_version == 'python-2.6.6' {
    $ld_lib_path = '/usr/lib64/python2.6'
  } else {
    $ld_lib_path = "/opt/rh/${python::python_version}/root/usr/lib64/"
  }
  file { '/etc/profile.d/set_ld_lib_path_for_python.sh':
    mode    => '0755',
    content => "export LD_LIBRARY_PATH=${ld_lib_path}\n",
  }
```

iakovgan:
correction: 

a better workaround of this issue might be :
```
  if $python::python_version == 'python-2.6.6' {
    $ld_lib_path = '/usr/lib64/python2.6'
  } else {
    $ld_lib_path = "/opt/rh/${python::python_version}/root/usr/lib64/"
  }
 file { "/etc/ld.so.conf.d/${python::python_version}.conf":
    content => "$ld_lib_path\n",
  }
```

stankevich:
Possibly fixed in https://github.com/stankevich/puppet-python/pull/314, please verify.

PR #314, opened 2016-07-15
 fix pip failing in virtualenv under SCL

mbmilligan:
When a virtualenv is created from an SCL python installation, that python binary needs to be runnable for commands in that virtualenv to work. At present this will fail if there is not a matching version of libpython installed in a system path. (Example: python3 is installed via SCL on a system without any version of python3 installed in the base system.) `scl enable` among other things, sets LD_LIBRARY_PATH, and thus needs to be invoked when running pip. `exec_prefix` should always be empty when SCL is not in use, so this should be a safe change.

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -117,7 +117,7 @@ define python::pip (
 
   $pip_env = $virtualenv ? {
     'system' => "${exec_prefix}pip",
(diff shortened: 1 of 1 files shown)


mbmilligan:
Possibly this is the issue in #234 as well

PR #323, opened 2016-08-22
Minor puppet-lint related fixes

cryptk:
In my use of this module, my CI system constantly complains about the two lines edited in this commit because they don't pass puppet-lint.  They are very minor, almost stylistic changes that I have made to the module internally (to appease the CI system) but it would be nice to not need to run a modified version of the module.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -18,7 +18,7 @@ class python::install {
   $python = $::python::version ? {
     'system' => 'python',
     'pypy'   => 'pypy',
-    default  => "${python::version}",
+    default  => $python::version,
   }
 
   $pythondev = $::osfamily ? {
@@ -220,8 +220,8 @@ class python::install {
     }
 
     package { 'gunicorn':
-      name   => $python::gunicorn_package_name,
       ensure => $gunicorn_ensure,
+      name   => $python::gunicorn_package_name,
     }
   }
 }


manifests/install.pp
```diff
@@ -18,7 +18,7 @@
   $python = $::python::version ? {
     'system' => 'python',
     'pypy'   => 'pypy',
-    default  => "${python::version}",
```
ghoneycutt: I think this is here to ensure that python::version is a string, by casting it as a string. Not sure if it is actually needed though.

manifests/install.pp
```diff
@@ -220,8 +220,8 @@
     }
 
     package { 'gunicorn':
-      name   => $python::gunicorn_package_name,
       ensure => $gunicorn_ensure,
+      name   => $python::gunicorn_package_name,
```
ghoneycutt: +1

ghoneycutt:
TravisCI is broken due to ruby gem issues. Created PR #324 to fix. Once it is merged, you will need to rebase against it. This will bring in those changes and hopefully allow your tests to pass.

manifests/install.pp
cryptk: Either way, that variable will still be an int or a str based on it's contents.  Quoting a variable will not cast it to a string.

```
root@testbox:~# cat test.pp 
$version=123
warning(type($version))
warning(type("${version}"))
```

```
```

cryptk:
No problems, I can rebase it after #324 is merged.

ghoneycutt:
Seems I have to fix the style along with another bug to get travis passing. Did that in my PR. Thank you for bringing attention to it!

I set puppet-lint to ignore the stringification of python::version. If you still want to solve it your way by not casting it into a string, please submit a new PR once #324 is merged. Just need to make sure it passes our tests.

Thanks again.

cryptk:
I think you missed my line comment.  Quoting the variable does not cast it to a string, it is an int either way.

PR #324, opened 2016-08-22
Fix travis

ghoneycutt:


manifests/install.pp
```diff
@@ -18,7 +18,7 @@
   $python = $::python::version ? {
     'system' => 'python',
     'pypy'   => 'pypy',
-    default  => "${python::version}",
+    default  => "${python::version}", # lint:ignore:only_variable_string
```
cryptk: This should be converted to just be default => $python::version rather than disabling the line message.  Placing the variable in quotes like this does not cast it to a string, and the behavior both ways is identical.  You would need to use sprintf to cast it to a string, but that would actually change the current behavior.

cryptk:
```
root@testbox:~# cat test.pp 
$version=123
warning(type($version))
warning(type("${version}"))
```

```
```

cryptk: And doing it another way just to demonstrate the "universality" of how puppet likes to put things into their little boxes:

```
root@testbox:~# cat test.pp 
$version_int=123
$version_quoted = "${version_int}"
warning(type($version_int))
warning(type($version_quoted))
```

```
Notice: Compiled catalog for cryptkcoding.com in environment production in 0.02 seconds
Notice: Finished catalog run in 0.30 seconds
```

In order to actually cast it into a string, you would need to do something weird like this:

```
root@testbox:~# cat test2.pp 
$version=123
$version_str = sprintf('"%s"', $version)
warning(type($version_str))
```

```
root@cryptkcoding:~# puppet apply test2.pp 
Warning: Scope(Class[main]): string
```

But again, doing that with sprintf would actually change the current behavior of the module as the current behavior is to treat the $python::version variable as either an int or a str depending on it's contents.  Removing the quotes and curly braces will resolve the lint issue while preserving that behavior.

ghoneycutt: @cryptk My only concern is that `$python` is stringified on purpose and by changing the current behavior it will break something that is expecting that `$python` is a string.

I'm in favor of doing it your way, though that needs to be tested. Please send a PR that changes the behavior.

cryptk: I think you misunderstood what I am saying.  Placing the variable in quotes like that does NOT stringify it.  It is still an integer (assuming that the input is just digits) or a string (if the input is not just digits) either way.

Changing it to read
```
default => $python::version
```
is behaviorally identical to the quoted and braced version, just more stylistically correct.  Please read my previous line comments which demonstrate this.

ghoneycutt:
Released in 1.14.1

PR #334, opened 2016-10-04
add an alias to the python-dev package

dannygoulder:
Allow other packages that require python-devel (e.g. on suse, redhat) to find the package in the catalog

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,10 +1,10 @@
 fixtures:
   repositories:
-    'stdlib':
-      repo: 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
+    stdlib:
+      repo: 'https://github.com/puppetlabs/puppetlabs-stdlib.git'
       ref: '4.6.0'
-    'epel':
-      repo: 'git://github.com/stahnma/puppet-module-epel.git'
+    epel:
+      repo: 'https://github.com/stahnma/puppet-module-epel.git'
       ref: '1.2.2'
   symlinks:
     python: "#{source_dir}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,81 +3,41 @@ language: ruby
 
 bundler_args: --without system_tests
 
-rvm:
-  - 1.8.7
-  - 1.9.3
-  - 2.0.0
-  - 2.1.0
-  - 2.3.1
-
-env:
-  matrix:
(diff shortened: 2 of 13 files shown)


ghoneycutt:
This is failing due to style. Please use two spaces instead of a hard tab, also the arrows must be aligned.

I don't think that adding an alias that is the same as the name is going to help identify the package. You can already reference this package resource by the title 'python-dev' and not by whatever $pythondev is.

dannygoulder:
Sorry, looks like something went wrong with tab conversion (I confess, I forked and edited it in the browser for speed). Should have linted locally before submitting the PR. :(

I am using another module (dwerder/graphite) that requires the 'python-devel' package on RHEL/CentOS - and expects it to be named as such (which seems reasonable). Because the package resource is named 'python-dev' puppet complains that 'python-devel' isn't in the catalog. I can't add 'python-devel' either as the namevar seems to make it collide. Perhaps arguably a shortcoming in Puppet, as maybe Puppet should alias the package name automatically.

Adding the alias fixes the problem by ensuring that the package resource will always match the actual package name in any require. It also seems to be pretty free of side effects.

I'll clean up the PR and resubmit it.

ghoneycutt:
Excellent, thank you!

Could you please update the spec tests to show the presence of this new attribute.

dannygoulder:
Sure, I'll try to get that done tomorrow. One thing I'm not sure about is whether adding an alias with the same name as the package will upset Puppet. Don't have a Debian-based box available to test it on though. Hopefully adding the tests will catch that one.

dannygoulder:
Just one thing to mention (I am just learning about spec tests so go easy on me :)). The spec tests are referencing Red Hat 5 (which is nearly EOL). Do you mind if I update that to 6 in the tests and update the PR? Or does it not really make much difference? Also just noticed that I missed adding the test for SuSe so I will add that too

dannygoulder:
Hmm looks like some incompatibility between a gem that got pulled in with rspec-puppet-facts and Ruby 1.8.7 causes those tests to fail. Newer ruby is working though. Might need some help with that... let me dig for a bit longer as I can use RHEL 6 ruby to fiddle.

dannygoulder:
Hi!  I abandoned the rspec-puppet-facts idea and I've rebased to master.

Issue #303, opened 2016-04-27
CentOS 7 with Python3 does not work


prozach:
Setting `python::version` to `3` as listed in the `params::valid_versions` doesn't work as the package `python3` cannot be found.  Looking in yum, only `python34` looks to be valid.  If I change the `python::version` to `34` (which is not listed as valid), python is installed, but I get a failure trying to install `python3-pip`, as the package does not exist in the the provided epel repo.  This seems to me like it would be a common use case for this module.  Any suggestions on how I should proceed?

mwasilew2:
yea, I've hit exactly the same issue, primary use case for me

wyardley:
+1
I'm also not totally sure why the validation doesn't fail as it is now, since 'python34' works, but doesn't appear to fall within the validation rules. I have a python34-pip available, however, setting `pip => present` doesn't ensure it; I have to manually add the python34-pip package resource in my own calling code.

Another problem I'm having is that when not using SCL, even after making sure that pip3 is available, I don't see an easy way (either in hiera or otherwise) to have it use a system pip (not SCL or virtualenv) with a different name, e.g., pip3; am I missing something?

wyardley:
ps - maybe it could have a pip3 provider (similar to puppet's), or some kind of passthrough, vs. trying to figure out whether to use 'pip' or 'pip3'?

zleswomp:
This has been implemented in the fork I'm using/maintaining.  https://github.com/xaque208/puppet-python

jameskirsop:
Unless I'm missing something, this issue still persists:

```
class { 'python' :
	version => '3',
	pip	=> 'present',
}
```
gives me a puppet run error:
`Error: /Stage[main]/Python::Install/Package[pip]/ensure: change from 'purged' to 'present' failed: Execution of '/bin/yum -d 0 -e 0 -y install 3' returned 1: Error: Nothing to do`

Changing `version => 3` to `version => python34` results in another puppet run error:
`Error: /Stage[main]/Python::Install/Package[pip]/ensure: change from 'purged' to 'present' failed: Execution of '/bin/yum -d 0 -e 0 -y install python3-pip' returned 1: Error: Nothing to do`

This is on CentOS7 with EPEL repos enabled.

Using the manifest example from the docs:
```
class { 'python' :
 ensure      => 'present',
 version     => 'rh-python36-python',
 dev         => 'present',
 virtualenv  => 'present',
}
```

gives me:
```
[...]

PR #347, opened 2016-12-10
Improve support for pip on CentOS7/EPEL

ju5t:
Due to a recent change in EPEL the python-pip package was renamed to python2-pip. This currently breaks this module on CentOS 7. This commit makes sure that it defaults to python2-pip if the system is running Python 2 on CentOS 7 when using EPEL.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -213,6 +213,9 @@ class python::install {
       if $::python::version =~ /^3/ {
         $pip_category = undef
         $pip_package = 'python3-pip'
+      } elsif ($::osfamily == 'RedHat') and (versioncmp($::operatingsystemmajrelease, '7') >= 0) and ($python::use_epel == true) {
+        $pip_category = undef
+        $pip_package = 'python2-pip'
       } elsif $::osfamily == 'Gentoo' {
         $pip_category = 'dev-python'
         $pip_package = 'pip'
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -213,7 +213,7 @@ class python::install {
       if $::python::version =~ /^3/ {
         $pip_category = undef
         $pip_package = 'python3-pip'
-      } elsif ($::osfamily == 'RedHat') and (versioncmp($::operatingsystemmajrelease, '7') >= 0) and ($python::use_epel == true) {
+      } elsif ($::osfamily == 'RedHat') and (versioncmp($::operatingsystemmajrelease, '7') >= 0) {
         $pip_category = undef
         $pip_package = 'python2-pip'
       } elsif $::osfamily == 'Gentoo' {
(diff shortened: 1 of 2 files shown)


manifests/install.pp
```diff
@@ -213,6 +213,9 @@
       if $::python::version =~ /^3/ {
         $pip_category = undef
         $pip_package = 'python3-pip'
+      } elsif ($::osfamily == 'RedHat') and (versioncmp($::operatingsystemmajrelease, '7') >= 0) and ($python::use_epel == true) {
```
aschaber1: I disabled use_epel since I have manage it elsewhere, but still would like this case to apply, which is why I like #349 better.

ju5t: I probably misinterpreted `$use_epel` when writing this. I'll remove it.

As far as #349 is concerned: it doesn't do any spec tests and uses a regex on a fact that can be done cheaper and more effectively with `versioncmp()` and `$::operatingsystemmajrelease`.

spec/classes/python_spec.rb
```diff
@@ -217,6 +217,22 @@
     end
   end
 
+  context "on a Redhat 7 OS" do
```
ghoneycutt: awesome work, thank you!

Would you mind adding a case for RedHat 6, that shows the same thing as this, but with the correct pip package name and update the existing RedHat 5 part with the name attribute. That would give us much better testing for all use cases with EL.

ghoneycutt:
@aschaber1 We need a third party to validate that this code actually works on EL7. Would you be able to do that and report back?

aschaber1:
@ghoneycutt, just tried it: works.
```
$ puppet agent -t --environment integration_python_pipfix
Info: Using configured environment 'integration_python_pipfix'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for $FQDN
Info: Applying configuration version '49315315c4582938518434a20cc6bacd3e9d38da'
Notice: /Stage[main]/Python::Install/Package[pip]/ensure: created
Notice: Applied catalog in 16.47 seconds
```
RPM Output:
```
# rpm -qa | grep pip
python2-pip-8.1.2-5.el7.noarch
```

ghoneycutt:
Thank you!

ghoneycutt:
Released in 1.18.2

wyardley:
I don't see python3-xxx (including python3-pip) available on CentOS7, even with epel repo enabled, though I do see python34-pip via our EPEL mirror. It's also not clear to me that this would help, even with python2-pip absent and python34-pip present, seems to me like the module is still hard-coded to use "pip".

Some additional comments in #303.

ju5t:
This pull request only solved an issue where packages were renamed by EPEL. What you're describing is a different issue. I think it would be better to discuss this in #303.

Issue #346, opened 2016-12-09
EPEL7 python-pip package is called python2-pip; puppet-python won't recognize that it is installed


talisto:
In EPEL7, the `python-pip` package is called `python2-pip`.  It is aliased to `python-pip`, so it can be installed with `yum install python-pip`, but then running `yum list python-pip` returns `No matching Packages`.  

This causes some problems when `puppet-python` tries to install the pip package:

```
Error: Execution of '/bin/yum -d 0 -e 0 -y list python-pip' returned 1: Error: No matching Packages to list
Error: /Stage[main]/Python::Install/Package[pip]/ensure: change from absent to present failed: Execution of '/bin/yum -d 0 -e 0 -y list python-pip' returned 1: Error: No matching Packages to list
Notice: /Stage[main]/Python/Anchor[python::end]: Dependency Package[pip] has failures: true
Warning: /Stage[main]/Python/Anchor[python::end]: Skipping because of failed dependencies
```

Is there a workaround for this?  I tried to figure out if there was a way that I could force `puppet-python` to look for `python2-pip` instead of `python-pip` but I couldn't see that as a configurable option.

I'm running CentOS 7.2.

talisto:
I fixed this by setting `python::version` to `python27` and `python::provider` to `scl`.

ghoneycutt:
Sounds like we need to change the name of the package.

PR #349, opened 2016-12-10
EL7 use python2-pip package name

### LINKED ISSUE #348: python-pip has been renamed to python2-pip on el7 epel repo
I found this week that the python-pip package is not available anymore for el7 on epel anymore, it has been replaced by python2-pip. That makes the module fail trying to check if it's installed as puppet does not really handle well the 'obsoletes' and similar rpm headers.

david-caro:
It turns out that the package name has changed in the epel repository.
closes #348

Signed-off-by: David Caro <david@dcaro.es>

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -216,6 +216,9 @@ class python::install {
       } elsif $::osfamily == 'Gentoo' {
         $pip_category = 'dev-python'
         $pip_package = 'pip'
+      } elsif ($::osfamily == 'RedHat') and ($::operatingsystemrelease =~ /^7/) {
+        $pip_category = undef
+        $pip_package = 'python2-pip'
       } else {
         $pip_category = undef
         $pip_package = 'python-pip'


schwicke:
This certainly would do the job. I wonder though if it would not be a good idea to allow the user to pass on the package name as a parameter. This way, it may be possible to simplify the code a bit.

david-caro:
I agree that is a good idea, and that it would have allowed me to work around the issue, though to me that should be the last resource for the user,it should pick the good package by default.

ghoneycutt:
Thank you for your contribution! Could you please add a spec test so that we can merge this change.

ghoneycutt:
@schwicke The package name should definitely be a parameter, though we also want to have sane defaults for our supported platforms so that when you use the module, it does the right thing out of the box without having to specify data.

ju5t:
You probably missed #346 when writing this pull request. Feel free to reuse the test I wrote. 

I would recommend `versioncmp()` over a regex on the wrong fact. You can use `$::operatingsystemmajrelease` instead.

ghoneycutt:
Hi,

Could someone please do a functional test of the code from https://github.com/stankevich/puppet-python/pull/347 and verify that it works to solve this problem?

ghoneycutt:
@david-caro Thanks for the contribution though going with PR #347 since it had spec tests. We could still use your help by verifying that the code in PR #347 works.

david-caro:
@ghoneycutt I did not see it, I think I searched on the issues but not the PRs. No problem 👍

schwicke:
@ghoneycutt fully agree, I think the  module should both provide sensible (working) defaults and the option to name the package as a parameter. That would be nice to have.

PR #355, opened 2016-12-23
Add python release as available facts

jcpunk:
For locating /usr/lib/pythonX.Y/ it is helpful to have access to just the specific release version.


ghoneycutt:
@jcpunk Thanks for adding these facts! Could you please provide tests as well. Here's an example

https://github.com/ghoneycutt/puppet-module-ssh/blob/master/spec/unit/facter/ssh_spec.rb

jcpunk:
I've taken a stab at tests based off the python_version tests

PR #358, opened 2017-01-25
Fix the package name for Amazon Linux

seefood:


wyardley:
I'm not sure if the test failures are due to this change or not, but I'm +1 on this change in general, we ran into the same problem.

may need to default `use_epel` to false too.

Shattoww:
Anyway we could get this upstream? Would help us as well.

PR #367, opened 2017-04-12
Removed duplicated warning

tux-o-matic:
Removed same check and warning of $virtualenv which was both on line 108 and 114


richburroughs:
Just saw this from looking at our logs and the code. It is checking that param and logging the error twice for no apparent reason.

tux-o-matic:
No merged PR in 6 months despite quite a few waiting to be.

bastelfreak:
Hi @tux-o-matic, we fixed this in our master branch.

PR #368, opened 2017-04-22
Add umask parameter to pip execs

jstaph:
I recently ran into an issue with a puppet server running under a [CIS](https://benchmarks.cisecurity.org/) hardened configuration, where the umask of the server was set rather aggressively.  Without the umask parameter, pip installations inherit the server's umask, which may, or may not, allow the installed egg to function correctly.

This change simply adds optional control of the umask to pip.pp


bastelfreak:
I rebased against master

PR #369, opened 2017-04-26
pip should be able to install modules with non PEP-440 versioning

vfoucault:
We should be able to install development modules with non PEP-440 versioning.
by default, installing a module with version => 0.0.1.dev0+d2da957 does not work.

I've added a dev_version parameter for pip, default to false witch permits to handle non PEP-440 version to be installed.

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -51,6 +51,9 @@
 # [*log_dir*]
 # String. Log directory.
 #
+# [*dev_version*]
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @vfoucault, can you please rebase?

bastelfreak:
Hi,
I'm going to close this PR due to inactivity. Please reopen and rebase it if you're still interested in it.

PR #378, opened 2017-08-28
Attempt to fix #357, ensure pip3 is used for py3.

### LINKED ISSUE #357: Issue "Can't install pip3 with Ubuntu #287" reemerged
The fix that resolved issue #287 has been broken again and now if you're using Ubuntu you can't install pip3 as the required value for version is "python3".  The cause seems to be the regex performed here: https://github.com/stankevich/puppet-python/blob/master/manifests/install.pp#L213

ram7os:
Tested on Ubuntu 14.04


ram7os:
wip: fixing the rspec tests

PR #385, opened 2017-11-17
Making sure, if you define a python binary, use it.

m2bright:
When you define a binary, use it.


m2bright:
Resolves: https://github.com/stankevich/puppet-python/issues/384

PR #388, opened 2018-04-21
fix for latest versions of setuptools and pip

vchepkov:
to properly initialize virtual environment pip needs to be upgraded before setuptools
pip version 10 obsoleted --no-use-wheel flag and replaced it with --no-binary :all:


vchepkov:
Also, instead of trying to guess, what parameters to pass give user an option to do it themselves

reist:
I just got bitten by the same. This fix works great!

PuppetNinja:
ditto, and got bitten by this also

PR #410, opened 2018-06-12
Fix Python version regex in install.pp

fklajn-opera:
According to the rest of the code `version` (if custom) is expected
to contain package name with version.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -215,7 +215,7 @@ class python::install {
         }
       }
 
-      if "${::python::version}" =~ /^3/ { #lint:ignore:only_variable_string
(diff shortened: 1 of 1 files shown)


manifests/install.pp
```diff
@@ -215,7 +215,7 @@
         }
       }
 
-      if "${::python::version}" =~ /^3/ { #lint:ignore:only_variable_string
+      if "${::python::version}" =~ /^python3/ { #lint:ignore:only_variable_string
```
wyardley: I see the lint ignore, but wonder if there's any reason this can't / shouldn't just be 
`if $python::version =~ /^python3/`

PR #412, opened 2018-07-07
Add support for providing pip provider

### LINKED ISSUE #303: CentOS 7 with Python3 does not work

danquack:
- Provide the ability to specify which pip provider to utilize. This will allow support for the use of pip3. Currently the syntax only specifies "pip" this will allow the execution to use "pip3"
- Updated readme to add documentation of how to install anything outside of system python

Fixes #303


README.md
```diff
@@ -77,6 +87,8 @@ Installs and manages packages from pip.
 
 **virtualenv** - virtualenv to run pip in. Default: system (no virtualenv)
 
+**pip_provider** - pip provder to execute pip with. Default: pip.
```
bastelfreak: there is a tiny typo: provder -> provider

manifests/pip.pp
```diff
@@ -66,11 +69,13 @@
 #
 # Sergey Stankevich
 # Fotis Gimian
+# Daniel Quackenbush
 #
 define python::pip (
   $pkgname         = $name,
   $ensure          = present,
   $virtualenv      = 'system',
+  $pip_proivder    = 'pip',
```
bastelfreak: Can you please add a dadatype for this new parameter? An enum would be awesome, otherwise String works as well.

manifests/requirements.pp
```diff
@@ -62,10 +62,12 @@
 # Sergey Stankevich
 # Ashley Penney
 # Fotis Gimian
+# Daniel Quackenbush
 #
 define python::requirements (
   $requirements           = $name,
   $virtualenv             = 'system',
+  $pip_provider           = 'pip',
```
bastelfreak: same here, please add a datatype

bastelfreak:
Hi @danquack, thanks for the PR! Can you please take a look at the used email address in your commit? It isn't associated with your github account. Can you also extend the current tests to validate variations of the new parameters?

danquack:
Thanks for the feedback. I fixed that typo real quick, and will work on tying/adding tests.

bastelfreak:
Hi @danquack, 

you can probably fix the rubocop errors with:

```
 bundle exec rake rubocop:auto_correct
```

danquack:
@bastelfreak copying over to a different fork to consolidate numerous git commits #414

PR #414, opened 2018-07-09
support for providing pip3 provider w/ tests. Modified readme 4 examples

### LINKED ISSUE #303: CentOS 7 with Python3 does not work

### LINKED ISSUE #377: How to deploy pip package to rhscl python34
On a RHEL server here we already have python34 from rhscl installed in /opt/rh/rh-python34
I'd like to install pip packages into this coming from a local mirror.
Can I do this using puppet-python and how?

(Btw: is this the correct place for questions? I did not find any other.)

danquack:

- Provide the ability to specify which pip provider to utilize. This will allow support for the use of pip3. Currently the syntax only specifies "pip" this will allow the execution to use "pip3"
- Updated readme to add documentation of how to install anything outside of system python
- Added rspec tests per @bastelfreak recommendations.

Fixes #303
Fixes #377

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Installs and manages python, python-pip, python-dev, python-virtualenv and Gunic
 
 **use_epel** - Boolean to determine if the epel class is used. Default: true on RHEL like systems, false otherwise
 
+*Install Python from system python*
 ```puppet
   class { 'python' :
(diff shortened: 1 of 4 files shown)


bastelfreak:
Thanks @danquack !

PR #429, opened 2018-09-06
Add single quotation for pip url

onstring:
Will fix the pip installation issue when url is `git+https://github.com/xxxxxxxx#egg=yyyy`

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -187,9 +187,9 @@ define python::pip (
 
   $source = $url ? {
     false               => "${pkgname}${extras_string}",
-    /^(\/|[a-zA-Z]\:)/  => $url,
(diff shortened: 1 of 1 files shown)


alexjfisher:
Does the regex 
https://github.com/voxpupuli/puppet-python/pull/429/files#diff-8a5246728de7306981fd3fa2191bd8eaR214 need adjusting?

danquack:
@onstring you can proabably fix the robocop errors with ``` bundle exec rake rubocop:auto_correct```

bastelfreak:
mhm, would an acceptance test be helpful here?

onstring:
The patch is fixing the breakage when url is like `git+https://github.com/xxxxxxxx#egg=yyyy`

alexjfisher:
and https://github.com/voxpupuli/puppet-python/pull/429/files#diff-8a5246728de7306981fd3fa2191bd8eaR214 ? This would have matched previously (when $source == $url), but won't anymore.

PR #434, opened 2018-10-05
Stop using 'pip search' for ensure => latest

### LINKED ISSUE #433: python::pip ensure => latest triggers refresh on each puppet run for some packages

- Puppet: 5.5.2
- Ruby: included in Puppet 5
- Distribution: Centos 7
- Module version: 2.1.0 

Try to use `python::pip` with `ensure => 'latest'` using https://github.com/devpi/devpi as PyPi mirror/own repo for packages that contain mixed case letters in their names.

## What are you seeing

Package is reinstalled on every puppet run and refresh is triggered, which causes unnecessary changes in reports and, what is worse, services restarts.

Package should not be updated if it's already in the latest version. No change should be reported for the resource and no refresh should be triggered.

Please see https://github.com/devpi/devpi/issues/592 for info how I tried to fix it on devpi side. Apart from getting to conclusion that this is hard to fix it there I realized that as pip team is considering to remove `pip search` altogether (see https://github.com/pypa/pip/issues/5216), perhaps even better solution for this problem would be to stop using pip search on puppet-python side. This could also fix issues like #350.

gdubicki:
Fixes #433

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -111,11 +111,6 @@ define python::pip (
       default => "--index-url=${index}",
     }
 
-  $pypi_search_index = $index ? {
-      false   => '',
-      default => "--index=${index}",
-    }
-
   $proxy_flag = $proxy ? {
     false    => '',
     default  => "--proxy=${proxy}",
@@ -243,10 +238,26 @@ define python::pip (
       }
 
       'latest': {
+        # Unfortunately this is the smartest way of getting the latest available package version with pip as of now
+        # Note: we DO need to repeat ourselves with "from version" in both grep and sed as on some systems pip returns
+        # more than one line with paretheses.
+        $latest_version = join(["${pip_env} install ${proxy_flag} ${pkgname}==notreallyaversion 2>&1",
+                                ' | grep -oP "\(from versions: .*\)" | sed -E "s/\(from versions: (.*?, )*(.*)\)/\2/g"',
+                                ' | tr -d "[:space:]"'])
+
+        # Packages with underscores in their names are listed with dashes in their place in `pip freeze` output
+        $pkgname_with_dashes = regsubst($pkgname, '_', '-', 'G')
+        $grep_regex_pkgname_with_dashes = "^${pkgname_with_dashes}=="
+        $installed_version = join(["${pip_env} freeze --all",
+                                  " | grep -i -e ${grep_regex_pkgname_with_dashes} | cut -d= -f3",
+                                  " | tr -d '[:space:]'"])
+
+        $unless_command = "[ \$(${latest_version}) = \$(${installed_version}) ]"
+
         # Latest version.
         exec { "pip_install_${name}":
           command     => "${wheel_check} ; { ${pip_install} --upgrade \$wheel_support_flag ${pip_common_args} || ${pip_install} --upgrade ${pip_common_args} ;}",
-          unless      => "${pip_env} search ${pypi_search_index} ${proxy_flag} ${source} | grep -i INSTALLED.*latest",
+          unless      => $unless_command,
           user        => $owner,
           group       => $group,
           umask       => $umask,
--- a/spec/defines/pip_spec.rb
+++ b/spec/defines/pip_spec.rb
@@ -64,22 +64,10 @@ describe 'python::pip', type: :define do # rubocop:disable RSpec/MultipleDescrib
 
         it { is_expected.not_to contain_exec('pip_install_rpyc').with_command(%r{--proxy}) }
       end
-      context 'does not add proxy to search command if set to latest and proxy is unset' do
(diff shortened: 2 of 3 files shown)


spec/defines/pip_spec.rb
```diff
@@ -64,22 +64,10 @@
...
       context 'adds proxy to install command if proxy set' do
         let(:params) { { proxy: 'http://my.proxy:3128' } }
 
-        it { is_expected.to contain_exec('pip_install_rpyc').with_command("pip wheel --help > /dev/null 2>&1 && { pip show wheel > /dev/null 2>&1 || wheel_support_flag='--no-binary :all:'; } ; { pip --log /tmp/pip.log install $wheel_support_flag  --proxy=http://my.proxy:3128   rpyc || pip --log /tmp/pip.log install  --proxy=http://my.proxy:3128   rpyc ;}") }
-      end
-      context 'adds proxy to search command if set to latest' do
-        let(:params) { { proxy: 'http://my.proxy:3128', ensure: 'latest' } }
-
-        it { is_expected.to contain_exec('pip_install_rpyc').with_command("pip wheel --help > /dev/null 2>&1 && { pip show wheel > /dev/null 2>&1 || wheel_support_flag='--no-binary :all:'; } ; { pip --log /tmp/pip.log install --upgrade $wheel_support_flag  --proxy=http://my.proxy:3128   rpyc || pip --log /tmp/pip.log install --upgrade  --proxy=http://my.proxy:3128   rpyc ;}") }
-        it { is_expected.to contain_exec('pip_install_rpyc').with_unless('pip search  --proxy=http://my.proxy:3128 rpyc | grep -i INSTALLED.*latest') }
```
danquack: I think this check also validated the proxy went through as defined in params.

gdubicki: Oh, right. I will restore this but without checking "unless". Would that be ok, @danquack ?

danquack: Yeah, as long as we’re not losing any validation checks.

gdubicki:
Note that this PR causes `$url` to be ignored when using `$ensure = latest` - only `$pkgname` is used.

I am not sure if that is wrong though because I doubt that the previous code did work correctly with `$url`..

What do you think, @danquack ?

manifests/pip.pp
```diff
@@ -271,10 +266,18 @@
       }
 
       'latest': {
+        # Unfortunately this is the smartest way of getting the latest available package version with pip as of now
+        $latest_version = join(["${pip_env} install ${proxy_flag} ${pkgname}==notreallyaversion 2>&1",
```
bastelfreak: We should really ensure that this works with an acceptance test.

gdubicki: Sure, I will work on that @bastelfreak .

bastelfreak:
@gdubicki can you check the failed travis tests?

gdubicki:
> @gdubicki can you check the failed travis tests?

I did, but I can't find the cause.. I am using Centos 7 in my infra and the code runs fine on many nodes for a few weeks now. I have spin up a vm with Debian 9 to test why `unless` seems not to work the way it should, but commands I am trying give me the same results as on Centos 7...

I don't really have much experience with tiny differences of Python-related stuff behaviour on various distros.. Do you have any ideas why it may be failing?

PS I will of course squash the commits at the end.

manifests/dotfile.pp
```diff
@@ -24,7 +24,7 @@
 #
 define python::dotfile (
   Enum['absent', 'present'] $ensure   = 'present',
-  Stdlib::Filemode $filename          = $title,
+  Any $filename                       = $title,
```
bastelfreak: mhm, what about this:
```suggestion
  String[1] $filename                 = $title,
```

gdubicki: I moved that fix to a separate PR: https://github.com/voxpupuli/puppet-python/pull/446

spec/defines/pip_spec.rb
gdubicki: In the end I removed those as we don't have "search command" anymore, so there is no point in checking if it contains proxy.

gdubicki:
Ready for final review, @bastelfreak @danquack .

ThiefMaster:
Wouldn't it be cleaner to use pip's JSON API and get the latest version from there? Check e.g. https://pypi.org/pypi/click/json - getting the "version" from there seems like a better option than relying on certain output when installing an invalid version..

gdubicki:
Can we merge this now, after https://github.com/voxpupuli/puppet-python/pull/446 got merged @bastelfreak @danquack ?

manifests/pip.pp
thaiphv: I think this doesn't fetch the packages from a custom index URL if someone happens to use it.

PR #440, opened 2018-10-21
Update modules with defined types for variables as described in docs/Add reference.md

danquack:
I couldn't find where you ran your merge script for updating the changelog, but I imagine an additional command of `puppet strings generate` can be run to make this automated and up to date.

#### Pull Request (PR) description
Based on the current docs, defined types for variables. Then executed rake strings to generate the reference markdown file for forge.

#### This Pull Request (PR) fixes the following issues
Fixes  #439
Fixes  #441

PR #443, opened 2018-10-26
Replace deprecated validate_* functions

baurmatt:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -61,11 +61,10 @@ class python (
     default => '',
   }
 
-  #$allowed_versions = concat(['system', 'pypy'], $valid_versions)
-  #unless $version =~ Enum[allowed_versions] {
-    #fail("version needs to be within${allowed_versions}")
-    #}
-  validate_re($version, concat(['system', 'pypy'], $valid_versions))
+  $allowed_versions = concat(['system', 'pypy'], $valid_versions)
+  unless $version =~ Enum[$allowed_versions] {
+    fail("version needs to be within${allowed_versions}")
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -95,11 +95,10 @@
     default => '',
   }
 
-  #$allowed_versions = concat(['system', 'pypy'], $valid_versions)
-  #unless $version =~ Enum[allowed_versions] {
-    #fail("version needs to be within${allowed_versions}")
-    #}
-  validate_re($version, concat(['system', 'pypy'], $valid_versions))
+  $allowed_versions = concat(['system', 'pypy'], $valid_versions)
+  unless $version =~ Enum[$allowed_versions] {
```
ekohl: Should this be `unless $version in $allowed_versions` if it's an array? Not sure what the data type of `$valid_versions` is.

baurmatt: `$valid_versions`, when taken from `python::params`, is an array. But even if it would be a string, the `concat()` would generate an array for `$allowed_versions`. I guess one could also use the `in` syntax you suggested. Not sure that the performance difference is.

baurmatt:
Is there a better way to verify data type errors then matching it with a Regex? Because for some reasons the output changes between Puppet 4,5 and 6....

wyardley:
Sorry, this may now have some conflicts now that I merged #440

baurmatt:
No worries! :) Rebased the branch.

bastelfreak:
Thanks for the PR @baurmatt !

PR #444, opened 2018-11-02
Fix for issue #303

### LINKED ISSUE #303: CentOS 7 with Python3 does not work

### LINKED ISSUE #442: Issues with CentOS7 and python3
This is a new issue to replace the prematurely closed #303 (as requested by @bastelfreak).

The core of this issue is that a number of combinations of configuration get python/virtualenv/pip3 partially installed, but not one single combination seems to get everything installed.

For example, providing the `version` of `python34` tries to install `python3-pip` which isn't a valid EPEL package name, and so there's a failure there. This is true for a number of the combinations listed in the original issue (#303).

To complicate things a little bit further, EPEL now has a `python36` package, but no matching `pip3` package - installing `pip3` requires `python34` to be installed from EPEL.

jameskirsop:
Adds better python34 pip support on CentOS7

Fixes #303 
Fixes #442

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -233,7 +233,7 @@ class python::install {
         }
       }
 
-      if "${::python::version}" =~ /^python3/ and ($::osfamily == 'RedHat') and (versioncmp($::operatingsystemmajrelease, '7') >= 0) { #lint:ignore:only_variable_string
+      if "${python::version}" =~ /^python3/ and ($facts['os']['family'] == 'RedHat') and (versioncmp($facts['os']['release']['major'], '7') >= 0) { #lint:ignore:only_variable_string
         $pip_category = undef
         $pip_package = 'python34-pip'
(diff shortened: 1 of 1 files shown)


bastelfreak:
@jameskirsop can you check the failed travis jobs?

manifests/install.pp
```diff
@@ -233,9 +233,12 @@
         }
       }
 
-      if "${::python::version}" =~ /^python3/ { #lint:ignore:only_variable_string
+      if "${python::version}" =~ /^python3/ and ($facts['os']['family'] == 'RedHat') and (versioncmp($facts['os']['release']['major'], '7') >= 0) { #lint:ignore:only_variable_string
```
wyardley: Does this not work with just:
`if $python::version`, which would fix the lint error and also avoid needing to ignore the lint error?

PR #446, opened 2018-11-04
Fix $filename and $mode types in python::dotfile

gdubicki:
erroneously changed in a990c0b

### DIFF
--- a/spec/defines/dotfile_spec.rb
+++ b/spec/defines/dotfile_spec.rb
@@ -1,0 +1,50 @@
+require 'spec_helper'
+
+describe 'python::dotfile', type: :define do
+  on_supported_os.each do |os, facts|
+    context("on #{os} ") do
+      let :facts do
+        facts
+      end
+
+      describe 'dotfile as' do
+        context 'fails with empty string filename' do
+          let(:title) { '' }
+
+          it { is_expected.to raise_error(%r{Evaluation Error: Empty string title at 0. Title strings must have a length greater than zero.}) }
+        end
+        context 'fails with incorrect mode' do
+          let(:title) { '/etc/pip.conf' }
+          let(:params) { { mode: 'not-a-mode' } }
+
+          it { is_expected.to raise_error(%r{Evaluation Error: Error while evaluating a Resource}) }
+        end
+        context 'succeeds with filename in existing path' do
+          let(:title) { '/etc/pip.conf' }
+
+          it { is_expected.to contain_file('/etc/pip.conf').with_mode('0644') }
+        end
+        context 'succeeds with filename in a non-existing path' do
+          let(:title) { '/home/someuser/.pip/pip.conf' }
+
+          it { is_expected.to contain_exec('create /home/someuser/.pip/pip.conf\'s parent dir').with_command('install -o root -g root -d /home/someuser/.pip') }
+          it { is_expected.to contain_file('/home/someuser/.pip/pip.conf').with_mode('0644') }
+        end
+        context 'succeeds when set owner' do
+          let(:title) { '/home/someuser/.pip/pip.conf' }
+          let(:params) { { owner: 'someuser' } }
+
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -5,9 +5,9 @@ root = true
 [*]
```
bastelfreak: We should probably deploy this via https://github.com/voxpupuli/modulesync_config/tree/master/moduleroot, are you interested to provide a patch to our modulesync_config repo? That would allow us to sync the `.editorconfig` to all of our repos.

spec/defines/dotfile_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe 'python::dotfile', type: :define do # rubocop:disable RSpec/MultipleDescribes
+
+  context 'on Debian OS' do
+    let :facts do
```
bastelfreak: can you please use [rspec-puppet-facts](https://github.com/mcanevet/rspec-puppet-facts) for this?

gdubicki: Yes.

.editorconfig
gdubicki: No, sorry but that's too much yak shaving for me. I will just revert it from this PR.

gdubicki:
Ready to review, @bastelfreak .

spec/defines/dotfile_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+          let(:title) { '/etc/pip.conf' }
+
+          it { is_expected.to contain_file('/etc/pip.conf').with_mode('0644') }
+        end
+        context 'succeeds with filename in a non-existing path' do
+          let(:title) { '/home/someuser/.pip/pip.conf' }
+
+          it { is_expected.to contain_exec('create /home/someuser/.pip/pip.conf\'s parent dir').with_command('install -o root -g root -d /home/someuser/.pip') }
+          it { is_expected.to contain_file('/home/someuser/.pip/pip.conf').with_mode('0644') }
+        end
```
danquack: Can I suggest adding two more tests to ensure complete coverage?
```        context 'succeeds when set owner' do
          let(:title) { '/home/someuser/.pip/pip.conf' }
          let(:params) { {'owner': 'someuser' } }

          it { is_expected.to contain_exec('create /home/someuser/.pip/pip.conf\'s parent dir').with_command('install -o someuser -g root -d /home/someuser/.pip') }
          it { is_expected.to contain_file('/home/someuser/.pip/pip.conf').with_owner('someuser') }
        end
        context 'succeeds when set group set' do
          let(:title) { '/home/someuser/.pip/pip.conf' }
          let(:params) { {'group': 'somegroup' } }

          it { is_expected.to contain_exec('create /home/someuser/.pip/pip.conf\'s parent dir').with_command('install -o root -g somegroup -d /home/someuser/.pip') }
          it { is_expected.to contain_file('/home/someuser/.pip/pip.conf').with_group('somegroup') }
        end
```

gdubicki:
It's failing on building image with Debian 9 and Puppet 6, long before testing my code. I propose to comment out this single setup from the matrix and re-enable it in a few days after it starts to work again (?) or fix it/keep it disabled if it wont (outside of this issue, of course).

ekohl:
Please don't comment those tests. They're temporary issues at Travis but still something we want to keep in. If it happens too often we should talk to Travis CI support about it.

gdubicki:
This broken test is blocking my PR from being merged. And this PR is fixing completely broken dotfile resource.

So I suggest we keep my commenting out that test and you can create an issue to fix it and create PR in it to re-enable the test when its fixed, @ekohl.

ekohl:
It was just an intermittent failure. It was likely that a restart of that particular job would have passed. That's why I strongly feel this PR shouldn't modify it.

bastelfreak:
Thanks @gdubicki !

danquack:
@bastelfreak - I believe this PR fixed #447, so that issue can probably be closed.

PR #450, opened 2018-11-16
Fix Ubuntu bionic package installation

ekohl:
This rewrites the OS-specific installation to a more generic case statement. It now only lists the trusty exception for the old naming which makes it future proof.

Replaces https://github.com/voxpupuli/puppet-python/pull/431


bastelfreak:
Thanks @ekohl !

PR #453, opened 2018-12-12
README.md: remove obsolete and redundant sections

kenyon:
Remove Installation and Compatibility sections from README.md. Puppet Forge has installation instructions on each module's page already. The Compatibility section is redundant with metadata.json.


bastelfreak:
Thanks for the cleanup @kenyon !

PR #461, opened 2019-01-15
Allow pip to work in AIX systems

feltra:
This changes allow installing python and pip modules, assuming that you have yum previously configured as per described at
https://www.ibm.com/developerworks/aix/library/aix-toolbox/date.html
It have been tested on AIX 6 and 7

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -12,7 +12,8 @@ class python::install {
     default  => "${python::version}", # lint:ignore:only_variable_string
   }
 
-  $pythondev = $::osfamily ? {
+  $pythondev = $facts['os']['family'] ? {
+    'AIX' => "${python}-devel",
     'RedHat' => "${python}-devel",
     'Debian' => "${python}-dev",
     'Suse'   => "${python}-devel",
@@ -83,7 +84,7 @@ class python::install {
       Package <| title == 'virtualenv' |> {
         name     => 'virtualenv',
         provider => 'pip',
-        require  => Package[$pythondev],
+        require  => Package['python-dev'],
       }
     }
     'scl': {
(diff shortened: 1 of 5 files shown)


manifests/install.pp
```diff
@@ -12,7 +12,8 @@
     default  => "${python::version}", # lint:ignore:only_variable_string
   }
 
-  $pythondev = $::osfamily ? {
+  $pythondev = $facts['os']['family'] ? {
+    'AIX' => "${python}-devel",
```
bastelfreak: can you please align the `=>`?

manifests/install.pp
```diff
@@ -177,17 +178,42 @@
       }
     }
     default: {
+      case $facts['os']['family'] {
+        'AIX': {
+          if "${::python::version}" =~ /^python3/ { #lint:ignore:only_variable_string
```
bastelfreak: please don't add leading `::`. That is only required for Puppet 3 compatibility.

manifests/params.pp
```diff
@@ -1,5 +1,6 @@
-# @api private
-# @summary The python Module default configuration settings.
+# == Class: python::params
```
bastelfreak: I think the old syntax was required for puppet-strings to correctly parse it.

manifests/params.pp
```diff
@@ -27,12 +29,17 @@
...
+    default => 'root'
+  }
+
   $pip_lookup_path = $facts['os']['family'] ? {
     'AIX' => [ '/bin', '/usr/bin', '/usr/local/bin', '/opt/freeware/bin/' ],
     default => [ '/bin', '/usr/bin', '/usr/local/bin' ]
   }
 
-  $gunicorn_package_name = $::osfamily ? {
+  $gunicorn_package_name = $facts['os']['family'] ? {
```
bastelfreak: Yey, :+1: for new hash facts \o/

manifests/pip.pp
```diff
@@ -69,10 +68,15 @@
...
   Array[String] $path                                        = ['/usr/local/bin','/usr/bin','/bin', '/usr/sbin'],
-) {
+){
   $python_provider = getparam(Class['python'], 'provider')
   $python_version  = getparam(Class['python'], 'version')
-
+  if $group {
+    $_group = $group
+  } else {
+    include ::python::params
```
bastelfreak: I think we can rely on getparam here as well? (like in line 72 and 73).

manifests/params.pp
feltra: 🍻

ghoneycutt:
Awesome work! Noticed that the `metadata.json` still needs to be updated. Something like the following.

```
    {
      "operatingsystem": "AIX",
      "operatingsystemrelease": [
        "6.1",
        "7.1",
        "7.2"
      ]
    }
```

feltra:
Finally =)

metadata.json
```diff
@@ -15,6 +15,13 @@
     "gunicorn"
   ],
   "operatingsystem_support": [
+     {
+      "operatingsystem": "AIX",
+      "operatingsystemrelease": [
+        "6100",
```
ghoneycutt: `facter operatingsystemrelease` returns `6100` instead of `6.1` ?

feltra: {name => AIX, family => AIX, release => {major => 6100-09-11-1810, full => 6100-09-11-1810}

feltra: I would keep this with family only and then handle the releases.

feltra: The return from operatingsystem release is 6100-09-11-1810..

bastelfreak: hrm this is pretty crappy from facter :(
@ghoneycutt do you think we should keep the current version of the metadata or update it to the actual value from facter?

feltra: I see for gentoo we don't care for release as well..

bastelfreak: We don't care for the versions on Gentoo and Archlinux because those are rolling release distributions. For everything else we have to care about it. Based on this data, the facts are mocked during the tests (and some other stuff).

bastelfreak:
@feltra can you take a look at the failing travis jobs?

bastelfreak:
thanks!

bastelfreak:
@feltra please also think about adding the facts to https://github.com/camptocamp/facterdb <3

PR #464, opened 2019-02-22
Add manage_scl boolean to control managing SCL

### LINKED ISSUE #463: Add switch to not manage SCL setup
Currently this module tries to manage the various SCL packages, which creates a few problems:

1) Duplicate package resources for `scl-utils` with any other module that might also try to manage SCL.
2) The `centos-release-scl` package contains the default repository configuration which is not always accessible in some enterprise environments so it is desirable to override the repositories with an internal mirror.

I propose adding a simple `manage_scl` boolean which can default to true for the purposes of maintaining current behaviour, but wraps the `centos-release-scl` and `scl-utils` packages so setting it to false will take these out of the catalog.

bodgit:
This adds a `manage_scl` boolean which can be set to `false` to stop managing the various SCL packages in the case some other module is doing that already.

Fixes #463

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -93,18 +93,26 @@ class python::install {
       # SCL is only valid in the RedHat family. If RHEL, package must be
       # enabled using the subscription manager outside of puppet. If CentOS,
       # the centos-release-SCL will install the repository.
-      $install_scl_repo_package = $::operatingsystem ? {
-        'CentOS' => 'present',
-        default  => 'absent',
-      }
+      if $python::manage_scl {
+        $install_scl_repo_package = $facts['os']['name'] ? {
+          'CentOS' => 'present',
+          default  => 'absent',
+        }
 
-      package { 'centos-release-scl':
-        ensure => $install_scl_repo_package,
-        before => Package['scl-utils'],
-      }
-      package { 'scl-utils':
-        ensure => 'latest',
-        before => Package['python'],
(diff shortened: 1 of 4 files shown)


manifests/install.pp
```diff
@@ -93,18 +93,26 @@
       # SCL is only valid in the RedHat family. If RHEL, package must be
       # enabled using the subscription manager outside of puppet. If CentOS,
       # the centos-release-SCL will install the repository.
-      $install_scl_repo_package = $::operatingsystem ? {
-        'CentOS' => 'present',
-        default  => 'absent',
-      }
+      if $python::manage_scl {
+        $install_scl_repo_package = $::operatingsystem ? {
```
bastelfreak: the topscope fact is deprecated. could you migrate it to `$facts['os']['name']`?

bodgit: Done.

bodgit:
Test failures appear unrelated to the change.

bastelfreak:
thanks!

bodgit:
If we can get a release with this change in, that would be awesome.

bastelfreak:
I'm working on a new release. That depends on https://github.com/voxpupuli/puppet-python/pull/467

bodgit:
Any chance of that release now that #467 has been merged?

bodgit:
Actually, scratch that for now, this module is still broken WRT to using SCL versions. I'll raise another issue.

PR #465, opened 2019-02-25
Revert "drop EOL operatingsystems"

thaiphv:
This reverts commit c9693a32cb2dbae3ec41c3a70c8035016d549836 as we still need to use the module on some of our servers that run on Precise.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -25,6 +25,7 @@
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
+        "5",
         "6",
         "7"
       ]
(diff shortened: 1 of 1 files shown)


Dan33l:
Ubuntu 12.04 is EOL since 28 04 2017.
We are in 2019.

bastelfreak:
hey @thaiphv. We only keep operatingsystems in the metadata.json that are not EOL and have tests. The systems you readded are all end of life so we cannot merge this. We're happy to help you migrating the platfrom to current operatingsystems. If you need some help feel free to join us on https://slack.puppet.com or our IRC channel #voxpupuli on freenode.

PR #468, opened 2019-03-25
Fix facts for python rc version

krissik:
Change regex in `python_version` and `python_release` facts so it matches python release candidate version (e.g. Python 2.7.16rc1, Python 3.7.3rc1).


bastelfreak:
@krissik can you please rebase and maybe add a test for this?

krissik:
Can someone point me to an example of how to test a fact?

krissik:
I will add tests but I do not know how. An example would be helpfull or some documentation.

igalic:
```irc
14:29 <meena> how do we rspec test facts? 
14:39 <ewoud> with pain
14:40 <ewoud> there's some boiler plate to clear facts before/after
14:42 <ewoud> https://github.com/voxpupuli/puppet-yum/blob/master/spec/unit/facter/yum_package_updates_spec.rb is probably a decent example
14:43 <ewoud> it is also one of the few examples that uses subject {}, most of them call Facter.fact() every time
14:43 <ewoud> maybe it should clear facts before rather than after though
14:44 <ewoud> it would be great if rspec-puppet provided some nice integration like it does with classes, defines, etc
14:48 <meena> yeah, that seems oddly asymmetric 
```

so, yeah, https://github.com/voxpupuli/puppet-yum/blob/master/spec/unit/facter/yum_package_updates_spec.rb

PR #472, opened 2019-04-03
Update version validation

### LINKED ISSUE #471: Unable to use SCL version

- Puppet: 5.x
- Ruby:
- Distribution: Red Hat
- Module version: 2.2.3-rc0

The following documented example doesn't work:
```puppet
class { 'python' :
  ensure     => 'present',
  version    => 'rh-python36-python',
  dev        => 'present',
  virtualenv => 'present',
}
```
## What are you seeing
This falls foul of the version validation:
```
version needs to be pypy, system or a version string like '3.5' or 'python3.5)
```

Should JFW

## Output log
See above

Previously in older versions of the module `validate_re()` was used along with the `$valid_versions` parameter however these patterns were not anchored so as long as `$version` contained a version number somewhere in the value, it would pass validation.

This example also won't work anyway with the validation fix, as the code will then try and install a `rh-python36-python-scldevel` package which doesn't exist, however there is a `rh-python36-scldevel` package. There's also a `rh-python36-python-devel` package which are the actual files to develop *against* Python, whereas the `-scldevel` package is for developing against the software collection itself. So I'm not sure the install class is pulling in the right package here anyway.

Simplest fix is to probably just update the regex with an additional pattern along the lines of `/\Arh-python\d{2}(?:-python)?\Z/` and that covers both cases.

### LINKED ISSUE #475: python3_version fact doesn't work on SCL

- Puppet: 5.x
- Ruby: 
- Distribution: Red Hat
- Module version: 2.2.3-rc0

Commits 72c34fabd1b3a2bf3e7f2a52c4bdd1ccd9b392cb and 3d102f4cf1d057aaaf13b47608d9724575ce2c9a have added the `python3_version` fact in the `python::pyvenv` defined type however this fact doesn't resolve to anything with SCL as the various binaries are not in `$PATH` by default.

## What are you seeing
Various errors along the lines of:
```
Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, 'versioncmp' parameter 'a' expects a String value, got Undef (file: /.../python/manifests/pyvenv.pp, line: 62, column: 9)
```

If I explicitly set the `version` parameter for `python::pyvenv` then I can skip over the first use of the fact here: https://github.com/voxpupuli/puppet-python/blob/e87db52002536e4f26da53c5000f90b21575668e/manifests/pyvenv.pp#L40

But then I still fall foul here: https://github.com/voxpupuli/puppet-python/blob/e87db52002536e4f26da53c5000f90b21575668e/manifests/pyvenv.pp#L62

It's probably better to use the computed/normalised version string in this call to `versioncmp()` rather than use the fact again.

`$::python::exec_prefix` is already set up correctly so after these checks complete the correct SCL-provided python binaries are used.

I'll add a fix for this to my existing #472 PR.

bodgit:
Update the version parameter validation to accept SCL-style Python versions, i.e. rh-python36

Fixes #471
Fixes #475

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -29,7 +29,7 @@
 # @example install python3 from scl repo
 #   class { 'python' :
 #     ensure      => 'present',
-#     version     => 'rh-python36',
+#     version     => 'rh-python36-python',
 #     dev         => 'present',
 #     virtualenv  => 'present',
 #   }
@@ -64,7 +64,7 @@ class python (
   }
 
   unless $version =~ Pattern[/\A(python)?[0-9](\.[0-9])+/,
-        /\Apypy\Z/, /\Asystem\Z/, /\Arh-python[0-9]{2}\Z/] {
+        /\Apypy\Z/, /\Asystem\Z/, /\Arh-python[0-9]{2}(?:-python)?\Z/] {
     fail("version needs to be pypy, system or a version string like '3.5' or 'python3.5)")
   }
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -29,7 +29,7 @@
 # @example install python3 from scl repo
 #   class { 'python' :
 #     ensure      => 'present',
-#     version     => 'rh-python36-python',
+#     version     => 'rh-python36',
 #     dev         => 'present',
 #     virtualenv  => 'present',
 #   }
@@ -64,7 +64,7 @@ class python (
   }
 
   unless $version =~ Pattern[/\A(python)?[0-9](\.[0-9])+/,
-        /\Apypy\Z/, /\Asystem\Z/] {
+        /\Apypy\Z/, /\Asystem\Z/, /\Arh-python[0-9]{2}\Z/] {
     fail("version needs to be pypy, system or a version string like '3.5' or 'python3.5)")
   }
 
--- a/spec/classes/python_spec.rb
+++ b/spec/classes/python_spec.rb
@@ -129,7 +129,7 @@ describe 'python', type: :class do
               end
             end
           end
-        when 'RedHat'
+        when 'RedHat', 'CentOS'
           case facts[:os]['release']['major']
           when '5'
             # written for RHEL 5
@@ -204,23 +204,6 @@ describe 'python', type: :class do
                   }
                 end
 
-                context 'scl' do
-                  describe 'with manage_scl' do
-                    context 'true' do
-                      let(:params) { { provider: 'scl', manage_scl: true } }
-
-                      it { is_expected.to contain_package('centos-release-scl') }
-                      it { is_expected.to contain_package('scl-utils') }
-                    end
-                    context 'false' do
-                      let(:params) { { provider: 'scl', manage_scl: false } }
-
-                      it { is_expected.not_to contain_package('centos-release-scl') }
-                      it { is_expected.not_to contain_package('scl-utils') }
-                    end
-                  end
-                end
-
                 # python::provider
                 context 'default' do
                   let(:params) { { provider: '' } }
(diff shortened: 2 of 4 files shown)


bastelfreak:
Hi @bodgit, thanks for the PR. Can you add some tests so we know that the issue is actually fixed?

bodgit:
@bastelfreak Fixed. I noticed a few problems with the tests:

1) It's keyed off `facts[:os]['name']` and only checks for Red Hat so CentOS is ignored completely. There are no Red Hat sample facts so basically none of these tests ever fire.
2) Most of the checks for Red Hat are within the EL5 context, and as only EL6+ is supported they're still not run.
3) SCL doesn't work on EL5 anyway so I've moved those tests specifically to the EL6 and EL7 contexts.

The tests now fail for on master and are fixed on this branch, so they're in a better state than they were.

spec/classes/python_spec.rb
```diff
@@ -262,12 +245,65 @@
...
               it { is_expected.to contain_class('python::install') }
               it { is_expected.to contain_package('pip').with_name('python-pip') }
+
+              describe 'with python::provider' do
+                context 'scl' do
+                  describe 'with version' do
+                    context '3.6 SCL package' do
+                      let(:params) { { version: 'rh-python36' } }
+
+                      it { is_expected.to compile }
```
bastelfreak: Can you update it like this:
```suggestion
                      it { is_expected.to compile.with_all_deps }
```

spec/classes/python_spec.rb
```diff
@@ -262,12 +245,65 @@
...
               it { is_expected.to contain_class('python::install') }
               it { is_expected.to contain_package('pip').with_name('python2-pip') }
+
+              describe 'with python::provider' do
+                context 'scl' do
+                  describe 'with version' do
+                    context '3.6 SCL package' do
+                      let(:params) { { version: 'rh-python36' } }
+
+                      it { is_expected.to compile }
```
bastelfreak:
```suggestion
                      it { is_expected.to compile.with_all_deps }
```

manifests/init.pp
```diff
@@ -29,7 +29,7 @@
 # @example install python3 from scl repo
 #   class { 'python' :
 #     ensure      => 'present',
-#     version     => 'rh-python36-python',
+#     version     => 'rh-python36',
```
bastelfreak: mhm, is `-python` not valid anymore? That would be a backwards-incompatible change.

bodgit: Either work to install the runtime as both packages basically require each other, but as I mentioned in #471 if you specify the version as `rh-python36-python` and elect to install the dev package, it will try and install `rh-python36-python-scldevel` which doesn't exist, yet `rh-python36-scldevel` does exist.

The original code has clearly been written with using `rh-python36` as the version as there is a commented out block that would have installed `${version}-python-virtualenv` which would also error if you tried to use `rh-python36-python` for the version, (too many `-python-`'s). It's been commented out as the `rh-python36` package will pull this in as a dependency, but `rh-python36-python` will not.

I'll update the regex so it doesn't break existing users but personally I think it needs a bit of work in the future; arguably the version _should_ actually be `rh-python36-python` and the dev option should install `rh-python36-python-devel` rather than `rh-python36-scldevel`, and `rh-python36-python-virtualenv` should be explicitly installed rather than be commented out like it is currently.

For reference, these are the "core" packages that are available for the 3.6 software collection:
```
rh-python36.x86_64 : Package that installs rh-python36
rh-python36-runtime.x86_64 : Package that handles rh-python36 Software Collection.
rh-python36-scldevel.x86_64 : Package shipping development files for rh-python36
rh-python36-python.x86_64 : Version 3 of the Python programming language aka Python 3000
rh-python36-python-debug.x86_64 : Debug version of the Python 3 runtime
rh-python36-python-devel.x86_64 : Libraries and header files needed for Python 3 development
rh-python36-python-setuptools.noarch : Easily build and distribute Python packages
rh-python36-python-virtualenv.noarch : Tool to create isolated Python environments
rh-python36-python-wheel.noarch : A built-package format for Python
```

bastelfreak:
Thanks @bodgit !

PR #474, opened 2019-04-07
modulesync 2.7.0

bastelfreak:
modulesync 2.7.0


dhoppe:
These tests must be converted from mocha to rspec-mocks.

PR #479, opened 2019-05-03
Dependency Bumps

danquack:
Just saw some dependencies that could use updating.


bastelfreak:
Hi @danquack. The Gemfile is templated for all of our modules based on https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/Gemfile.erb and https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml. Updating rubocop would cause a lot of work to some modules. Are you interested to work on the rubocop issues that this would cause on the other modules?

PR #469, opened 2019-03-26
Allow http proxy on bootstrap

feltra:

Issue #473, opened 2019-04-05
missing https_proxy when using https pypi of other https indexes


jsfrerot:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4
- Distribution: centos7
- Module version: 2.2.2

## How to reproduce
install a python3 virtualenv without direct internet access. But use proxy to download packages from pypi
```
python::virtualenv { '/opt/myproject' :
        ensure     => present,
        distribute => false,
        version    => '36',
        owner      => $system_user,
        group      => 'myuser',
        cwd        => '/opt/myproject',
        proxy      => 'http://my_proxy:3128',
        virtualenv => 'virtualenv-3',
        require    => User['myuser'],
    }
```

## What are you seeing
a timeout when installing/upgrading the pip package

## What behaviour did you expect instead
a clean install

## Here is the diff to fix the issue:
```
diff --git a/modules/python/manifests/virtualenv.pp b/modules/python/manifests/virtualenv.pp
index ad526678f..8e99e8d81 100644
--- a/modules/python/manifests/virtualenv.pp
+++ b/modules/python/manifests/virtualenv.pp
@@ -125,7 +125,7 @@ define python::virtualenv (
 
     $proxy_command = $proxy ? {
       false   => '',
-      default => "&& export http_proxy=${proxy}",
+      default => "&& export http_proxy=${proxy} https_proxy=${proxy}",
     }
 
     # Virtualenv versions prior to 1.7 do not support the
```

PR #488, opened 2019-05-22
Allow HTTP_PROXY on bootstrap.

pillarsdotnet:
Fixes: #469 #473

### DIFF
--- a/manifests/pip/bootstrap.pp
+++ b/manifests/pip/bootstrap.pp
@@ -3,6 +3,7 @@
 #
 # @param version should be pip or pip3
 # @param manage_python if python module will manage deps
+# @param http_proxy Proxy server to use for outbound connections.
 #
 # @example 
 #   class { 'python::pip::bootstrap':
@@ -11,6 +12,7 @@
 class python::pip::bootstrap (
   Enum['pip', 'pip3'] $version            = 'pip',
   Variant[Boolean, String] $manage_python = false,
+  Optional[Stdlib::HTTPUrl] $http_proxy   = undef,
 ) inherits ::python::params {
   if $manage_python {
     include python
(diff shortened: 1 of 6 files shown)


manifests/pip/bootstrap.pp
```diff
@@ -19,26 +20,34 @@
...
-        require => Package['python3'],
+        command     => '/usr/bin/curl https://bootstrap.pypa.io/get-pip.py | python3',
+        environment => $environ,
+        unless      => 'which pip3',
+        path        => $python::params::pip_lookup_path,
+        require     => Package['python3'],
       }
       # puppet is opinionated about the pip command name
-      file { 'pip3-python':
+      -> file { 'pip3-python':
```
bastelfreak: please remove the arrow or the `require` parameter. Only one of it is needed.

pillarsdotnet: Done.

manifests/pip/bootstrap.pp
```diff
@@ -11,6 +11,7 @@
 class python::pip::bootstrap (
   Enum['pip', 'pip3'] $version            = 'pip',
   Variant[Boolean, String] $manage_python = false,
+  String $http_proxy                      = '',
```
Dan33l: Missing doc about this new parameter and then update of `REFERENCE.md`.

pillarsdotnet: Done.

manifests/pip/bootstrap.pp
```diff
@@ -19,12 +21,19 @@
       'AIX' => '/opt/freeware/bin',
       default => '/usr/bin'
     }
+
+    $environ = $facts['os']['family'] ? {
+      'AIX' => [ "http_proxy=${http_proxy}", "https_proxy=${http_proxy}" ],
+      default => [ "HTTP_PROXY=${http_proxy}", "HTTPS_PROXY=${http_proxy}" ]
```
alexjfisher: I wonder if some users have already been setting these environment variables outside of this module.  This would unset them.  Maybe best to only set `environment` if an http_proxy has been specified.
```puppet
if $http_proxy { # Assuming we change the default to `undef`
  $environ = $facts['os']['family'] ? {
    'AIX'   => [ "http_proxy=${http_proxy}", "https_proxy=${http_proxy}" ],
    default => [ "HTTP_PROXY=${http_proxy}", "HTTPS_PROXY=${http_proxy}" ],
  }
} else {
  $environ = undef
}
```

manifests/pip/bootstrap.pp
```diff
@@ -3,6 +3,7 @@
 #
 # @param version should be pip or pip3
 # @param manage_python if python module will manage deps
+# @apram http_proxy Proxy server to use for outbound connections.
```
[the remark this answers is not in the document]
pillarsdotnet: blah!

Fixing...

manifests/pip/bootstrap.pp
```diff
@@ -11,6 +12,7 @@
 class python::pip::bootstrap (
   Enum['pip', 'pip3'] $version            = 'pip',
   Variant[Boolean, String] $manage_python = false,
+  String $http_proxy                      = '',
```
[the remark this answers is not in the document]
pillarsdotnet: @alexjfisher 
Okay, but note that I'm following the existing coding pattern.  If I started cleaning things up to my normal standards, we'd wind up changing nearly every line.

manifests/pip/bootstrap.pp
pillarsdotnet: Done.

pillarsdotnet:
@bastelfreak @alexjfisher @Dan33l -- I'm done making changes, and tests passed.

alexjfisher:
@pillarsdotnet Thanks, it's looking good.  Are you able to add any tests?

pillarsdotnet:
@alexjfisher   I could add rspec tests maybe, but I'm not sure they would add any value.  Acceptance tests would require setting up a proxy-server, and that's a little too advanced for me.

manifests/virtualenv.pp
```diff
@@ -134,12 +137,12 @@
...
 
     exec { "python_virtualenv_${venv_dir}":
-      command     => "true ${proxy_command} && ${virtualenv_cmd} ${system_pkgs_flag} -p ${python} ${venv_dir} && ${pip_cmd} --log ${venv_dir}/pip.log install ${pip_flags} --upgrade pip && ${pip_cmd} install ${pip_flags} --upgrade ${distribute_pkg}",
+      command     => "${virtualenv_cmd} ${system_pkgs_flag} -p ${python} ${venv_dir} && ${pip_cmd} --log ${venv_dir}/pip.log install ${pip_flags} --upgrade pip && ${pip_cmd} install ${pip_flags} --upgrade ${distribute_pkg}",
       user        => $owner,
       creates     => "${venv_dir}/bin/activate",
       path        => $_path,
       cwd         => '/tmp',
-      environment => $environment,
+      environment => $environment.union($proxy_environ),
```
pillarsdotnet: What would happen if `$proxy_environ` and `$environment` contained different values for `HTTP_PROXY`/`HTTPS_PROXY`?  How would we write a test to find out?

alexjfisher: Writing the tests shouldn't be too complicated.  You can set any combination of `environment` and `proxy` parameters and then assert the `environment` of the `exec`.  eg. (completely untested, incomplete and probably contains typos etc!)

```ruby
context 'when `proxy` set' do
  let(:title) { '/opt/env1' }
  let(:params) { { proxy: 'http://example.com:3128' } }
  it { is_expected.to contain_exec('python_virtualenv_/opt/env1').with_environment(['HTTP_PROXY=http://example.com:3128','HTTPS_PROXY=http://example.com:3128']) }
end

context 'when `proxy` and `environment` have conflicting parameters' do
  let(:title) { '/opt/env1' }
  let(:params) do
    {
      proxy: 'http://example.com:3128',
      environment: [ 'HTTP_PROXY=http://example.com:8080' ]
    }
  end
  it { is_expected.to contain_exec('python_virtualenv_/opt/env1').with_environment(['HTTP_PROXY=http://example.com:3128','HTTPS_PROXY=http://example.com:3128']) }
end
# etc.
```

I think that second example will fail though.  `union` will not deduplicate the conflicting `HTTP_PROXY=` strings.  If you can add tests like this, I can help with a fix for replacing `union` (assuming the test *does* fail).  Hope this helps.

pillarsdotnet: Well, `union` on't deduplicate environment strings, but presumably the shell will.

pillarsdotnet: Test added; let's see if it passes.

pillarsdotnet:
Yay! Tests passed!

bastelfreak:
Thanks @pillarsdotnet !

PR #489, opened 2019-05-27
Fix version-check.

### LINKED ISSUE #303: CentOS 7 with Python3 does not work

### LINKED ISSUE #354: Python 3 + virtualenv + centos 7 not working
Hi

I've been trying to set up a virtualenv on a centos 7.2 using python 3. No luck so far, basically python3-virtualenv doesn't exist, so no virtualenv. 
I tried to say `python { ... virtualenv => absent ...}` then `python::pip { 'virtualenv': }` but I end up with a dependency cycle due to Exec[pip_install_virtualenv]. 

My suggestion would be to use os package for python, python-dev and python-pip then stick to pip for virutalenv and other. But maybe there's another reason for not doing so ?

Best regards,

### LINKED ISSUE #420: Support for Python3.6 executables
Just leaving a question here.
Am currently experiencing setting up python+virtualenv using python3.6
Have installed it through the epel repositories, which has given the executable /bin/python36

Now setting up virtual environments through python3.6 no longer uses the virtualenv pip package, instead it can be set up using the command /bin/python36 -m venv /path/to/virtualenv.

Here's what I have experienced so far:
- python::virtualenv does not work because of the exec resource it tries to invoke
- python::pip, including the virtualenv parameter does not make any change at all

Is there any implementation if the latest version of this module to support it?
Just leaving a question here.

Operating system used - CentOS7
Puppet version - Puppet4
Module version used - v2.1.0 (used the tag)

### LINKED ISSUE #476: Allow the use of pip3.4 and pip3.6
Centos 7 doesn't install pip3 anymore, but pip3.4 or pip3.6

```
diff --git a/modules/python/manifests/pip.pp b/modules/python/manifests/pip.pp
index e7aa6efe2..3c8f241b9 100644
--- a/modules/python/manifests/pip.pp
+++ b/modules/python/manifests/pip.pp
@@ -75,7 +75,7 @@ define python::pip (
   $pkgname                             = $name,
   $ensure                              = present,
   $virtualenv                          = 'system',
-  Enum['pip', 'pip3'] $pip_provider    = 'pip',
+  Enum['pip','pip3','pip3.4','pip3.6'] $pip_provider    = 'pip',
   $url                                 = false,
   $owner                               = 'root',
   $group                               = 'root',
```

pillarsdotnet:
Fixes #303
Fixes #354
Fixes #420
Fixes #429 
Fixes #444
Fixes #476

Allows Python3 installation on Centos-7 thusly:

```
class { 'python':
  version: 'python36'
}
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -65,9 +65,9 @@ class python (
     default => '',
   }
 
-  unless $version =~ Pattern[/\A(python)?[0-9](\.[0-9])+/,
+  unless $version =~ Pattern[/\A(python)?[0-9](\.?[0-9])+/,
         /\Apypy\Z/, /\Asystem\Z/, /\Arh-python[0-9]{2}(?:-python)?\Z/] {
-    fail("version needs to be pypy, system or a version string like '3.5' or 'python3.5)")
+    fail("version needs to be pypy, system or a version string like '36', '3.6' or 'python3.6' )")
   }
 
   # Module compatibility check
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -6,13 +6,12 @@
 #
 class python::install {
 
-  $python_version = getparam(Class['python'], 'version')
-  $python = $python_version ? {
-    'system'                        => 'python',
-    'pypy'                          => 'pypy',
-    /\A(python)?([0-9](\.?[0-9])+)/ => "python${2}",
-    /\Arh-python[0-9]{2}/           => $python_version,
-    default                         => "python${python::version}",
+  $python = $python::version ? {
+    'system'              => 'python',
+    'pypy'                => 'pypy',
+    /\A(python)?([0-9]+)/ => "python${2}",
+    /\Arh-python[0-9]{2}/ => $python::version,
+    default               => "python${python::version}",
   }
 
   $pythondev = $facts['os']['family'] ? {
@@ -57,6 +56,7 @@ class python::install {
 
   package { 'virtualenv':
     ensure  => $venv_ensure,
+    name    => "${python}-virtualenv",
     require => Package['python'],
   }
 
@@ -144,10 +144,10 @@ class python::install {
     }
     'rhscl': {
       # rhscl is RedHat SCLs from softwarecollections.org
-      if $::python::rhscl_use_public_repository {
-        $scl_package = "rhscl-${::python::version}-epel-${::operatingsystemmajrelease}-${::architecture}"
+      if $python::rhscl_use_public_repository {
+        $scl_package = "rhscl-${python::version}-epel-${::operatingsystemmajrelease}-${::architecture}"
         package { $scl_package:
-          source   => "https://www.softwarecollections.org/en/scls/rhscl/${::python::version}/epel-${::operatingsystemmajrelease}-${::architecture}/download/${scl_package}.noarch.rpm",
+          source   => "https://www.softwarecollections.org/en/scls/rhscl/${python::version}/epel-${::operatingsystemmajrelease}-${::architecture}/download/${scl_package}.noarch.rpm",
           provider => 'rpm',
           tag      => 'python-scl-repo',
         }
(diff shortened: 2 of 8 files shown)


manifests/init.pp
```diff
@@ -65,7 +65,7 @@
     default => '',
   }
 
-  unless $version =~ Pattern[/\A(python)?[0-9](\.[0-9])+/,
+  unless $version =~ Pattern[/\A(python)?[0-9](\.?[0-9])+/,
         /\Apypy\Z/, /\Asystem\Z/, /\Arh-python[0-9]{2}(?:-python)?\Z/] {
     fail("version needs to be pypy, system or a version string like '3.5' or 'python3.5)")
```
rnelson0: Now that this now allows python35, perhaps update the fail message?

manifests/install.pp
```diff
@@ -199,8 +199,8 @@
     default: {
       case $facts['os']['family'] {
         'AIX': {
-          if "${python_version}" =~ /^python3/ { #lint:ignore:only_variable_string
-            class { 'python::pip::bootstap':
+          if String($::python::version) =~ /^python3/ {
```
rnelson0: This and the String on L271 should not have :: prepends, or the uses above should have them - just for consistency.

manifests/install.pp
```diff
@@ -268,25 +268,34 @@
...
         $pip_category = undef
         $pip_package = 'python2-pip'
-      } elsif $::osfamily == 'Gentoo' {
-        $pip_category = 'dev-python'
-        $pip_package = 'pip'
+        $pip_provider = pip2
       } else {
-        $pip_category = undef
-        $pip_package = 'python-pip'
+        if $::osfamily == 'Gentoo' {
```
rnelson0: Can this be kept as an elsif block, like the others? The nesting inside the final else block is a little confusing.

pillarsdotnet: It could, but then the two assignments get repeated unnecessarily.

manifests/install.pp
pillarsdotnet: Adding prepends everywhere; I'm not sure which style is preferred.

manifests/init.pp
pillarsdotnet: Updated.

pillarsdotnet:
> The code looks functional, a few comments on layout/syntax. This would benefit from an additional test or two leveraging the new version string allowed, though.

Made requested changes; added tests; waiting for them to pass.

pillarsdotnet:
@rnelson0 -- Woo-hoo!  Tests passed.

manifests/install.pp
```diff
@@ -6,13 +6,12 @@
...
-    'pypy'                          => 'pypy',
-    /\A(python)?([0-9](\.?[0-9])+)/ => "python${2}",
-    /\Arh-python[0-9]{2}/           => $python_version,
-    default                         => "python${python::version}",
+  $python = $::python::version ? {
+    'system'              => 'python',
+    'pypy'                => 'pypy',
+    /\A(python)?([0-9]+)/ => "python${2}",
+    /\Arh-python[0-9]{2}/ => $::python::version,
+    default               => "python${::python::version}",
```
bastelfreak: can you please undo the leading `::`? From our [review guidelines](https://voxpupuli.org/docs/#reviewing-a-module-pr):
> In the majority of cases, variables shouldn’t be accessed via topscope: `$::modulename::$param`. Instead do: `$modulename::$param`

pillarsdotnet:
Removed `::` prefixes as requested.

Yes, I'm using this locally to install requisites for [pre-commit](https://pre-commit.com/).

manifests/install.pp
pillarsdotnet: Done.

jozsi:
This is actually breaking. If I specify version `3.7` or `python3.7` (because the Ubuntu package is called `python3.7`), it matches the `/\A(python)?([0-9]+)/` regular expression and the 2nd group matches only `3`, so the `.7` gets cut off in the package version.

bastelfreak:
@jozsi thanks for pointing this out. Are you able to provide a PR with a test that matches this?

jozsi:
@bastelfreak - I am away for a couple of days. I just used the latest
"puppet-agent-ubuntu" image (now based on Ubuntu 18.04 Bionic) and
provision with a puppet manifest that only contains puppet-python set to
version "python3.7" or "3.7". Both gets interpreted as "python3" under the
hood, that is already satisfied by "python3.6" installed on the system, so
no action is taken by ensure.

PR #490, opened 2019-05-30
Fix $subscribe overloading

### LINKED ISSUE #470: Variable $subscribe shoud not be overwritten

- Puppet: 4.8.2
- Ruby: 2.3.3
- Distribution: Debian GNU/Linux 9.5 (stretch)
- Module version: 2.2.2

```puppet
python::requirements { '/tmp/requirements.txt':
    virtualenv => $venv_directory,
    subscribe  => Vcsrepo[$app_directory],
    require    => Python::Pyvenv[$venv],
}
```

## What are you seeing 

I cannot make any of my instances of `python::requirements`  subscribe to other resources because inside of puppet-python `requirements.pp` the `$subscribe` variable (meta variable of puppet) is overwritten. 

I'd expect to be able to subscribe to a resource. Renaming `$subscribe` inside `requirements.pp` to something else would probably be enough. 

Because of that I had to keep `stankevich-python` instead of updating it to voxpupuli version.

## Output log
```log
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Cannot reassign variable '$subscribe' at /etc/puppet/modules/python/manifests/requirements.pp:134:16 at /etc/puppet/modules/profile/manifests/api.pp:151 on node 2950a0e0-flower.2950a0e0
```

nward:

In python::requirements, we simply rename $subscribe to $local_subscribe, to prevent it being redefined. This is a puppet metaparameter so should not be used as a variable within in a define.

Fixes #470

### DIFF
--- a/manifests/requirements.pp
+++ b/manifests/requirements.pp
@@ -89,9 +89,9 @@ define python::requirements (
       replace => false,
       content => '# Puppet will install and/or update pip packages listed here',
     }
-    $subscribe = File[$requirements]
+    $local_subscribe = File[$requirements]
   } else {
-    $subscribe = undef
+    $local_subscribe = undef
   }
 
   exec { "python_requirements${name}":
@@ -101,7 +101,7 @@ define python::requirements (
     timeout     => $timeout,
     cwd         => $cwd,
     user        => $owner,
-    subscribe   => $subscribe,
+    subscribe   => $local_subscribe,
     environment => $environment,
   }
 }


bastelfreak:
Thanks for the fix @nward !

manifests/requirements.pp
```diff
@@ -101,7 +101,7 @@
     timeout     => $timeout,
     cwd         => $cwd,
     user        => $owner,
-    subscribe   => $subscribe,
+    subscribe   => $local_subscribe,
```
tonythomas01: Correct me if Im wrong, it sounds impossible to set subscribe when I want `$manage_requirements == false`.

PR #491, opened 2019-06-09
Fix python::pip installing $editable VCS packages every Puppet run

### LINKED ISSUE #193: Installing from git repo runs install on every Puppet run
$ pip install -e git+https://github.com/dcramer/piplint.git#egg=piplint-dev
$ pip freeze | grep piplint
-e git://github.com/dcramer/piplint.git@c1c7921784d0315b659e0b53a2752ee5f0d202fe#egg=piplint-dev

The current check use regex: ^piplint==, which correlates to:
piplint==0.1.1

And as such, the module determines that piplint isn't installed, and starts install again

wolttam:
This is due to VCS packages appearing in `pip freeze -all` as:
-e git://example.com/package.git@<ref>#egg=<egg>

Updates $grep_regex, which checks whether a given version is installed, to
match against the two different output formats of `pip list`. No longer
relies on pip freeze.

Additionally, this commit:

Removes the multiple exec resources which were mostly the same and replaces
with variables $pip_exec_name, $command, and $unless_command, which get
used in a single exec resource at the end of the manifest.

Checks if $pkgname contains ==, and if so splits based on == and sets
$real_pkgname to the first part, and $ensure variable to the second part.
Uses of $pkgname have been replaced with $real_pkgname.

Fixes #193

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -135,14 +135,23 @@ define python::pip (
     fail('python::pip cannot provide uninstall_args with ensure => present')
   }
 
-  # Check if searching by explicit version.
-  if $ensure =~ /^((19|20)[0-9][0-9]-(0[1-9]|1[1-2])-([0-2][1-9]|3[0-1])|[0-9]+\.\w+\+?\w*(\.\w+)*)$/ {
-    $grep_regex = "^${pkgname}==${ensure}\$"
-  } else {
-    $grep_regex = $pkgname ? {
-      /==/    => "^${pkgname}\$",
-      default => "^${pkgname}==",
+  if $pkgname =~ /==/ {
+    $parts = split($pkgname, '==')
+    $real_pkgname = $parts[0]
+    $_ensure = $ensure ? {
+      'absent' => 'absent',
+      default => $parts[1],
     }
+  } else {
+    $real_pkgname = $pkgname
+    $_ensure = $ensure
+  }
+
+  # Check if searching by explicit version.
+  if $_ensure =~ /^((19|20)[0-9][0-9]-(0[1-9]|1[1-2])-([0-2][1-9]|3[0-1])|[0-9]+\.\w+\+?\w*(\.\w+)*)$/ {
+    $grep_regex = "^${real_pkgname}[[:space:]]\\+(\\?${_ensure}\\()$\\|$\\|, \\|[[:space:]]\\)"
+  } else {
+    $grep_regex = "^${real_pkgname}[[:space:]].*$"
   }
 
   $extras_string = empty($extras) ? {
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @wolttam, thanks for the PR. Can you add an acceptance test to verify that the bug is fixed? Please also have a look at the failed travis jobs.

wolttam:
I fixed what was causing the existing tests to fail -- missing quotes (oops).

And, I added a new test for this, except, here's a quandary... git (or some other vcs, but git seems to be the appropriate choice) is required for it to run. I'm quite unfamiliar with the test framework in use here, so I'm not sure how to make git available. I'm looking into it, though.

bastelfreak:
this will install git within the docker environment:

```patch
diff --git a/spec/acceptance/virtualenv_spec.rb b/spec/acceptance/virtualenv_spec.rb
index a3520c3..dbcfd6d 100644
--- a/spec/acceptance/virtualenv_spec.rb
+++ b/spec/acceptance/virtualenv_spec.rb
@@ -113,21 +113,22 @@ describe 'python class' do
     end
     it 'works with editable=>true' do
       pp = <<-EOS
-      class { 'python' :
+      package{'git':
+        ensure => 'present',
+      }
+      -> class { 'python' :
         version    => 'system',
         pip        => 'present',
         virtualenv => 'present',
       }
-      ->
-      python::virtualenv { 'venv' :
+      -> python::virtualenv { 'venv' :
         ensure     => 'present',
         systempkgs => false,
         venv_dir   => '/opt/venv5',
         owner      => 'root',
         group      => 'root',
       }
-      ->
-      python::pip { 'rpyc' :
+      -> python::pip { 'rpyc' :
         ensure     => '4.1.0',
         url        => 'git+https://github.com/tomerfiliba/rpyc.git',
         editable   => true,

```

wolttam:
> this will install git within the docker environment (and improves the formatting):
> ....
Of course! Added that, pushing now.

manifests/pip.pp
```diff
@@ -135,14 +135,21 @@
     fail('python::pip cannot provide uninstall_args with ensure => present')
   }
 
+  $real_pkgname = $pkgname
+  if $pkgname =~ /==/ {
+    $parts = split($pkgname, '==')
+    $real_pkgname = $parts[0]
```
bastelfreak: you reassign `$real_pkgname`. This should not work. The assignment in line 128 should probably happen in an else block. Can you also add a test where the pkgname contains a `==`?

wolttam: Made this fix, and added the test.

spec/acceptance/virtualenv_spec.rb
```diff
@@ -111,5 +103,57 @@ class { 'python' :
...
+      }
+      EOS
+
+      # Run it twice and test for idempotency
+      apply_manifest(pp, catch_failures: true)
+      apply_manifest(pp, catch_changes: true)
+    end
+    it 'works with == in pkgname' do
+      pp = <<-EOS
+      'python' :
```
bastelfreak: I think something got lost here?

```suggestion
      class{'python' :
```

wolttam: Ouch! While I'm add it, would it be preferred if I removed the leading space before all resource names?

Also, I found what's causing the first two tests to fail: using $path in the exec instead of $_path, going to push that along with this correction.

bastelfreak: Removing the leading space isn't required and I think it would just make the diff harder to read. I think we can keep that as is.

bastelfreak:
Thanks for the awesome work!

PR #497, opened 2019-06-18
Fix a reassigned variable

### LINKED ISSUE #496: Python::Pip fails if $ensure='absent'

- Puppet: 6.4.2
- Ruby: 
- Distribution: Ubuntu
- Module version: 3.0.1

```
python::pip { 'PROGRAMM':
  ensure       => 'absent',
  pkgname      => 'PROGRAMM',
  pip_provider => 'pip',
  virtualenv   => '/opt/PROGRAMM',
}
```

## What are you seeing
```
error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Cannot reassign variable '$pip_exec_install'
```
This occurres in `modules/python/manifests/pip.pp` in line 224, column 27

A clean run without an error.

SaschaDoering:
This occurred when $ensure was set to 'absent'. Then the variable used
for the title of the exec resource was assigned twice, resulting in an
error.

I solved it via a variable to avoid having to set the title in several places.
Of course, I can also put the title of the exec resource into another variable if the syntax there looks too weird. Please let me know if I should do this.

Fixes #496

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -174,8 +174,6 @@ define python::pip (
   $pip_install = "${pip_env} --log ${log}/pip.log install"
   $pip_common_args = "${pypi_index} ${proxy_flag} ${install_args} ${install_editable} ${source}"
 
-  $pip_exec_name = "pip_install_${name}"
-
   # Explicit version out of VCS when PIP supported URL is provided
   if $source =~ /^'(git\+|hg\+|bzr\+|svn\+)(http|https|ssh|svn|sftp|ftp|lp|git)(:\/\/).+'$/ {
     if $_ensure != present and $_ensure != latest {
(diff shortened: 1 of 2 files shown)


manifests/pip.pp
```diff
@@ -221,14 +219,17 @@
...
         $unless_command = "! ${pip_env} list | grep -i -e '${grep_regex}'"
       }
     }
   }
 
-  exec { $pip_exec_name:
+  exec { (defined('$uninstall')) ? {
+      true  => "pip_uninstall_${name}",
+      false => "pip_install_${name}",
+    }:
```
baurmatt: Can you please do this outside of the resource title? This will help to make it better readable.

manifests/pip.pp
```diff
@@ -221,14 +219,19 @@
 
       default: {
         # Anti-action, uninstall.
-        $pip_exec_name = "pip_uninstall_${name}"
+        $uninstall = true
```
baurmatt: Do we need a extra variable? Can't we just use $ensure?

SaschaDoering: No. Yes. I`ll implement that.

PR #498, opened 2019-06-21
Use shell to exec pip commands by default

jamebus:
Use shell to exec pip commands by default

Sometimes Puppet will auto-select the POSIX provider to run exec resources. This will use the shell provider by default for pip execs where it matters. Needed for pipes and the bang operator.


bastelfreak:
thansk for the fix @jamebus !

PR #500, opened 2019-07-03
Add option for not managing python and virtualenv packages.

tdukaric:

Helps in cases where other modules are already managing those packages,
avoids duplicate resource definition issues.


spec/classes/python_spec.rb
```diff
@@ -31,6 +31,18 @@
             end
           end
 
+          describe 'with python::manage_python_package' do
```
bastelfreak: can you also add a test that explicitly sets the new parameters to false and check if the catalog still compiles and the packages won't be in it?

PR #510, opened 2019-10-09
modulesync 2.9.0

bastelfreak:
modulesync 2.9.0


bastelfreak:
14.04 doesn't work properly with pypi anymore, also it's EOL. We are dropping it.

bastelfreak:
I've the strong feeling that we should rip out python 2.7 support :(. Anybody around that wants to look into the failed acceptance tests?

PR #513, opened 2019-10-25
Collect python sub-resources from all hiera levels

IATkachenko:
Hiera may be used at module, environment and global levels.
Default lookup strategy is "first", so pips, virtualenvs, pyvenvs,
requirements and dotfiles will be collected just from one level and in
case of two independent modules wants some pips (for example), only one pip will
be chosen for installing.
To avoid it we should use deep merge strategy in lookup function.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -46,11 +46,6 @@ class python (
   $gunicorn_package_name                          = $python::params::gunicorn_package_name,
   Optional[Enum['pip', 'scl', 'rhscl', 'anaconda', '']] $provider = $python::params::provider,
   $valid_versions                                 = $python::params::valid_versions,
-  Hash $python_pips                               = { },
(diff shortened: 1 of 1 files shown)


baurmatt:
I prefer not do a deep merge in a module because there might be people who don't need it and they cannot simply change this option.

I would suggest to use `lookup_options` directly in Hiera --> https://puppet.com/docs/puppet/5.5/hiera_merging.html#concept-2997

IATkachenko:
> I prefer not do a deep merge in a module because there might be people who don't need it and they cannot simply change this option.
> 
> I would suggest to use `lookup_options` directly in Hiera --> https://puppet.com/docs/puppet/5.5/hiera_merging.html#concept-2997

I see.
Thank you for explanation and sorry for disturbance.

PR #523, opened 2019-12-05
Allow python::version to contain a point (e.g. python3.7)

baurmatt:
#### Pull Request (PR) description
This worked before #489.

#### This Pull Request (PR) fixes the following issues
Fixes #508.

PR #524, opened 2019-12-05
Allow minor version in python::version e.g. python3.7

b4ldr:


b4ldr:
Seems the user also got help on slack and they beat me to the PR https://github.com/voxpupuli/puppet-python/pull/523

PR #537, opened 2020-02-27
Fixes for virtualenv_version fact when virtualenv > 20.x

### LINKED ISSUE #534: Virtualenv>=20.0.0 has new version command output breaking virtualenv_version fact

We are running old everything, but the facter fact, https://github.com/voxpupuli/puppet-python/blob/master/lib/facter/virtualenv_version.rb, hasn't been changed in a while.

Try to use `virtualenv_version` fact when `virtualenv>=20.0.0` is installed

## What are you seeing

```
error while resolving custom fact "virtualenv_version": undefined method `[]' for nil:NilClass
```

`virtualenv_version` is defined with a version value.

This happens because `virtualenv>=20.0.0` is a major rewrite and changes the output of the virtualenv --version command.

Old virtualenv:
```
$ virtualenv --version
16.7.5
```

New virtualenv:
```
$ ~/venv/bin/virtualenv --version
virtualenv 20.0.1 from /home/clark/venv/lib/python2.7/site-packages/virtualenv/__init__.pyc
```

The fact's output parsing will need to be updated to handle both cases.

pjonesIDBS:

### DIFF
--- a/lib/facter/virtualenv_version.rb
+++ b/lib/facter/virtualenv_version.rb
@@ -3,7 +3,7 @@
 Facter.add('virtualenv_version') do
   setcode do
     if Facter::Util::Resolution.which('virtualenv')
-      Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(%r{^(\d+\.\d+\.?\d*).*$})[0]
+      Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(%r{(\d+\.\d+\.?\d*).*$})[1]
     end
   end
 end
--- a/spec/unit/facter/virtualenv_version_spec.rb
+++ b/spec/unit/facter/virtualenv_version_spec.rb
@@ -5,19 +5,44 @@ describe Facter::Util::Fact do
     Facter.clear
   end
 
-  let(:virtualenv_version_output) do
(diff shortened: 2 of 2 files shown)


saz:
Looks good to me.

Failing tests seems to be related to `setuptools` and deprecation of python 2

bastelfreak:
@pjonesIDBS I rebased your PR and it should be green now. However, can you add a test for it?

pjonesIDBS:
Hi Tim (@bastelfreak),

Just cloned my fork of the module and inspected it in VSC. It looks like there's already accetance tests in spec/acceptance/facts_test_spec.rb. There's also a unit test in spec/unit/facter/virtualenv_version_spec.rb. 

I converted the module locally to use PDK (which worked like a charm).

The easiest solution to test was to update the unit test for virtualenv_version.

I set the factor "virtualenv_version_output" to be what we see with a version > 20.x (in this case 20.0.17) and then validated the facter would equal 20.0.17 with the rest of the string removed.

All looks good. The builds are still passing as they did previously.

saz:
@pjonesIDBS I think it makes sense to check for both kinds of version, as this module might still be used with older virtualenv versions

PR #542, opened 2020-04-06
Require puppet-epel over stahnma-epel

dhoppe:


bastelfreak:
duplicate of https://github.com/voxpupuli/puppet-python/pull/545

PR #550, opened 2020-06-08
fix site-packages parameter for focal fossa

crazymind1337:
On Ubuntu 20.04 there is no parameter `--no-site-packages` for `virtualenv` binary anymore. The `--system-site-packages` stays the same.

Fixes '--no-site-packages' parameter on Ubuntu Focal Fossa

```
root@6e6956e915f1:/# grep -i focal /etc/os-release 
VERSION="20.04 LTS (Focal Fossa)"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

```
root@6e6956e915f1:/# virtualenv --version
virtualenv 20.0.17 from /usr/lib/python3/dist-packages/virtualenv/__init__.py
```

```
root@6e6956e915f1:/# virtualenv --help | grep 'no-site-packages'
root@6e6956e915f1:/# 
```

```
root@6e6956e915f1:/# virtualenv --help | grep '^[\ -]*system-site-packages'
  --system-site-packages        give the virtual environment access to the system site-packages dir (default: False)
```

### DIFF
--- a/manifests/virtualenv.pp
+++ b/manifests/virtualenv.pp
@@ -97,11 +97,18 @@ define python::virtualenv (
     if (( versioncmp($_virtualenv_version,'1.7') > 0 ) and ( $systempkgs == true )) {
       $system_pkgs_flag = '--system-site-packages'
     } elsif (( versioncmp($_virtualenv_version,'1.7') < 0 ) and ( $systempkgs == false )) {
-      $system_pkgs_flag = '--no-site-packages'
+      if $facts['os']['release']['full'] == '20.04' {
+        $system_pkgs_flag = ''
+      } else {
+        $system_pkgs_flag = '--no-site-packages'
+      }
     } else {
       $system_pkgs_flag = $systempkgs ? {
         true    => '--system-site-packages',
-        false   => '--no-site-packages',
+        false   => $facts['os']['release']['full'] ? {
+          '20.04' => '',
+          default => '--no-site-packages'
+        },
         default => fail('Invalid value for systempkgs. Boolean value is expected')
       }
     }


igalic:
and this doesn't affect any other Debian type OS?

crazymind1337:
> and this doesn't affect any other Debian type OS?

Ah you are right. Just did check for our needs. I will check also for other OSes later this day. Thanks for reminding!

crazymind1337:
@igalic I checked other supported OSes (using latest docker container). At Debian the `virtualenv` binary is not used and on latest CentOS `virtualenv` does still have the `--no-site-packages` option.

manifests/virtualenv.pp
```diff
@@ -97,11 +97,18 @@
     if (( versioncmp($_virtualenv_version,'1.7') > 0 ) and ( $systempkgs == true )) {
       $system_pkgs_flag = '--system-site-packages'
     } elsif (( versioncmp($_virtualenv_version,'1.7') < 0 ) and ( $systempkgs == false )) {
-      $system_pkgs_flag = '--no-site-packages'
+      if $facts['os']['release']['full'] == '20.04' {
```
logicminds, line 100: I might suggest that this conditional logic here put into moduledata instead.  This is only a suggestion to clean this up though.

crazymind1337: There is no module hieradata yet.

logicminds: Would be easy to add if you have the time.

crazymind1337: Sure this is a good idea, but I am currently too busy with adding other 20.04 support. And it is also not part of the PR to change the params.pp to module hieradata. Ths should be done in another PR/Ticket.

saz:
I've missed this PR and made a new one: https://github.com/voxpupuli/puppet-python/pull/552

I think it doesn't make sense to check for any OS version, as `--no-site-packages` is deprecated since a long time.
(see https://virtualenv.pypa.io/en/16.7.9/reference.html)
The default has changed to not use site-packages anyways, so we don't need to be explicit here.

crazymind1337:
> I've missed this PR and made a new one: #552
> 
> I think it doesn't make sense to check for any OS version, as `--no-site-packages` is deprecated since a long time.
> (see https://virtualenv.pypa.io/en/16.7.9/reference.html)
> The default has changed to not use site-packages anyways, so we don't need to be explicit here.

So i can close this one?

saz:
@crazymind1337 Good question.

@igalic What do you think?

igalic:
i think we've accumulated a lot more feedback on what the best approach to solving this issue is in @saz's pr, so let's keep going there

PR #551, opened 2020-07-10
Fix virtualenv cmd

saz:
`--no-site-packages` argument is deprecated and can't be used with newer versions of virtualenv


igalic:
hmm?

PR #555, opened 2020-07-21
Fix undefined method error in facts

### LINKED ISSUE #527: "undefined method []" in  pip_version and virtualenv_version facts
We ran into a situation where `pip` and `virtualenv` were in the PATH on a particular system but referred to a non-existent python installation.

In both cases the facts are generated by using a regex match against the stdout/stderr of the relevant commands.  When the regex fails to produce a match it returns `nil` instead and of course `nil` doesn't have a `[]` method.

- Puppet: 6.9.0
- Ruby: 2.5.3p105
- Distribution: CentOS 7
- Module version: 3.0.1

Run puppet with the `python` module on a system with broken `pip` and/or `virtualenv` commands.

## What are you seeing

Errors:

```
Error: Facter: error while resolving custom fact "pip_version": undefined method []' for nil:NilClass
Error: Facter: error while resolving custom fact "virtualenv_version": undefined method []' for nil:NilClass
```

The `pip_version` and `virtualenv_version` facts should be `nil` when their respective commands are broken.

wiebe:
Prevents fact errors when pip/virtualenv installation is broken and adds version fact for pip2 and pip3.

Fixes #527

### DIFF
--- a/lib/facter/pip_version.rb
+++ b/lib/facter/pip_version.rb
@@ -1,9 +1,26 @@
 # Make pip version available as a fact
 
+def get_pip_version(executable)
+  if Facter::Util::Resolution.which(executable) # rubocop:disable Style/GuardClause
+    results = Facter::Util::Resolution.exec("#{executable} --version 2>&1").match(%r{^pip (\d+\.\d+\.?\d*).*$})
+    results[1] if results
+  end
+end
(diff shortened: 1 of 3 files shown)


bastelfreak:
thanks for the PR!

PR #557, opened 2020-07-31
Update virtualenv_version.rb

### LINKED ISSUE #534: Virtualenv>=20.0.0 has new version command output breaking virtualenv_version fact

GusPriceAtlassian:

### DIFF
--- a/lib/facter/virtualenv_version.rb
+++ b/lib/facter/virtualenv_version.rb
@@ -3,7 +3,7 @@
 Facter.add('virtualenv_version') do
   setcode do
     if Facter::Util::Resolution.which('virtualenv')
-      Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(%r{(\d+\.\d+\.?\d*).*$})[1]
+      Facter::Util::Resolution.exec('virtualenv --version 2>&1').match(%r{(\d+\.\d+\.?\d*).*$})[0]
     end
   end
 end


pjonesIDBS:
Tested on my local RHEL 7 server the proposed change would reintroduce the previous issues with virtualenv > 20.x:
[root@puppet facter]# facter -p virtualenv_version
20.0.29 from /usr/lib/python2.7/site-packages/virtualenv/__init__.pyc

The current code results in the desired output:
[root@puppet facter]# facter -p virtualenv_version
20.0.29

Versions older than 20.x will continue to work normally as before:
[root@puppet facter]# facter -p virtualenv_version
15.1.0

This was produced using Python 2.7 but I don't believe it should matter which version of Python is installed...

[root@puppet ~]# file /bin/virtualenv
/bin/virtualenv: C++ source, ASCII text

The Puppet fact is simply running this binary and trying to parse the output. It's odd then if we're seeing different results to the Ruby match regex on different systems. Certainly this fix has worked for others as well.

GusPriceAtlassian:
It's funny, I just ran it all again on my machine in AWS (CentOS 7 marketplace AMI) and got the right result using your Ruby. I think I may have just been in too deep at the end of last week trying to get this to work. But, it's working now, so I will take it! Sorry for the bother, thanks for your help.

pjonesIDBS:
Hi Gus, no problem! It happens to us all :) Pleased to see another person actively contributing. Glad we don't have a big mystery to delve into this time too. Have a good week.

PR #563, opened 2020-09-12
Drop EOL Debian 8

bastelfreak:


ghoneycutt:
Should drop EL6 too instead of figuring out the errors in PR #564

ghoneycutt (approved):

PR #564, opened 2020-09-12
CentOS 6: Enable acceptance tests

bastelfreak:
our metadata.json claims to support CentOS 6. In that case we should
also test on it.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -6,6 +6,7 @@
     - set: ubuntu1804-64
     - set: debian8-64
     - set: debian9-64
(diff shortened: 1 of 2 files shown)


bastelfreak:
* CentOS 6 tests don't work anymore (probably since a very long time)
* CentOS 6 is soon EOL
* CentOS 6 ships python 2.6 which is EOL since ages
* We already support CentOS 7

I'm :+1: to drop CentOS 6 unless someone else fixes the tests.

PR #566, opened 2020-09-13
Drop CentOS 6 support

bastelfreak:
We don't run acceptance tests for CentOS 6 since a long time.
In https://github.com/voxpupuli/puppet-python/pull/564 we tried to
enable the acceptance tests again but they don't work.

* CentOS 6 ships Python 2.6 which is EOL since ages
* CentOS 6 is EOL by end of the year anyway
* We already support CentOS 7


igalic:
it's one thing to not run acceptance tests for it, but an entirely different beast to drop support before the EOL date

bastelfreak:
I dont know if this even still works on CentOS 6. the tests arent present since a long time. And I think as a community we should not encourage people to use outdated software. And we should not claim CentOS 6 support in the metadata.json if we know that the tests dont pass. If anyone wants to fix them, thats fine by me. But I also think we ahould drop support as soon as it is end of life (which we also did in the past). The EoL date fot CentOS 6 is soon.

ghoneycutt:
I mentioned we should drop EL6 in https://github.com/voxpupuli/puppet-python/pull/563 though didn't realize it was not fully EOL yet. Agree with @igalic that we should not deprecate before the actual EOL.

Also agree with @bastelfreak that if the tests do not actually pass, we have two options, stop saying we support something we do not, or do the work to support it. I'm OK with either and suggest we merge this. This does not preclude anyone from properly adding EL6 support.

ghoneycutt (approved):

PR #576, opened 2020-10-05
Implement Ubuntu 20.04 support

bastelfreak:

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -284,8 +284,11 @@ class python::install {
         $pip_package  = 'pip'
         $pip_provider = 'pip'
       } else {
+        $pip_package = fact('os.release.major') ? {
+          '20.04' => 'python3-pip',
+          default => 'python-pip',
+        }
         $pip_category = undef
-        $pip_package  = 'python-pip'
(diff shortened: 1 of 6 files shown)


bastelfreak:
I think before we implement this properly we should drop the python::virtualenv resource as it's really really end of life. otherwise we need to add a huge amount of workarounds.

DnZmfr:
Hi,
Do you have any idea when this fix for ubuntu 20.04 will be released?

Thanks.
Dan

kenyon:
> Do you have any idea when this fix for ubuntu 20.04 will be released?

Not before the tests pass, at least.

bastelfreak:
@DnZmfr I'm a bit low on available time since I also maintain other modules. If you're interested in picking this up or dropping python::virtualenv in another PR, that would be highly appreciated :)

treydock:
@bastelfreak I would think it's not possible to get rid of `python::virtualenv` unless it's okay to force RHEL 7 sites to use Python 3.  Right now that defined type is the only way to use system python and virtualenv on RHEL7 and still works on RHEL8 because the OS provides Python 2.7 packages.

PR #578, opened 2020-10-31
Rework determination of virtualenv site-packages flags for v20

PeterJCLaw:
`virtualenv` 20.0 removed `--no-site-packages` so we cannot pass that on that version. This means we have three version scenarios to cope with:

* `< 1.7`: site packages installed by default, with `--no-site-packages` to opt out

* `1.7` - `20.0`: `--no-site-packages` present and default, `--system-site-packages` to opt back in

* `> 20.0`: no site packages by default, `--system-site-packages` to opt back in

This PR aims to add support for the last of these, so that all three are supported.

This was mentioned at https://github.com/voxpupuli/puppet-python/issues/534#issuecomment-585869365 and I've hit this myself (on Fedora 33, where this patch fixes it for me).

Meta: I've not yet added any tests; I'm not sure how to do that, but if you can point me in the right direction I'm happy to give it a go.

### DIFF
--- a/manifests/virtualenv.pp
+++ b/manifests/virtualenv.pp
@@ -86,24 +86,27 @@ define python::virtualenv (
       }
     }
 
-    # Virtualenv versions prior to 1.7 do not support the
-    # --system-site-packages flag, default off for prior versions
-    # Prior to version 1.7 the default was equal to --system-site-packages
-    # and the flag --no-site-packages had to be passed to do the opposite
+    # Virtualenv handling of site packages has changed over time, both its
+    # default and the flags to control it. See inline comments for details.
     $_virtualenv_version = getvar('virtualenv_version') ? {
       /.*/ => getvar('virtualenv_version'),
       default => '',
     }
-    if (( versioncmp($_virtualenv_version,'1.7') > 0 ) and ( $systempkgs == true )) {
-      $system_pkgs_flag = '--system-site-packages'
-    } elsif (( versioncmp($_virtualenv_version,'1.7') < 0 ) and ( $systempkgs == false )) {
-      $system_pkgs_flag = '--no-site-packages'
-    } else {
-      $system_pkgs_flag = $systempkgs ? {
-        true    => '--system-site-packages',
-        false   => '--no-site-packages',
-        default => fail('Invalid value for systempkgs. Boolean value is expected')
-      }
+    $system_pkgs_flag = $systempkgs ? {
+      # --system-site-packages was introducd in 1.7, and --no-site-packages
+      # became the default
+      true    => versioncmp($_virtualenv_version,'1.7') > 0 ? {
+        true  => '--system-site-packages',
+        # Rely on default prior to 1.7
+        false => '',
+      },
+      # --no-site-packages was removed in 20.0
+      false   => versioncmp($_virtualenv_version,'20') < 0 ? {
+        true  => '--no-site-packages',
+        # Rely on default after 20.0
+        false => '',
+      },
+      default => fail('Invalid value for systempkgs. Boolean value is expected')
     }
 
     $distribute_pkg = $distribute ? {


manifests/virtualenv.pp
```diff
@@ -86,24 +86,27 @@
...
-    } else {
-      $system_pkgs_flag = $systempkgs ? {
-        true    => '--system-site-packages',
-        false   => '--no-site-packages',
-        default => fail('Invalid value for systempkgs. Boolean value is expected')
-      }
+    $system_pkgs_flag = $systempkgs ? {
+      # --system-site-packages was introducd in 1.7, and --no-site-packages
+      # became the default
+      true    => versioncmp($_virtualenv_version,'1.7') > 0 ? {
```
PeterJCLaw, line 98: Note: I thiiink that this should probably be `>=`, however I believe the current spelling has the same error.

crazymind1337:
I have tested your branch at our CI. The result is what I expected:

The CI fails due to the fact $virtualenv is not set, because it is not installed for a fresh system. Therefore the comparison which you are making it not catching up.

My PR which was based on the $OS, like: if $os == 20.04 => remove --no-site-packages parameter has not been accepted due to the fact the removal of the parameter is based on the virtualenv version. 

I think we will have to modify the exec for `virtualenv`. Maybe we make 2 execs with an `onlyif` parameter which differs on the different virtualenv version.

crazymind1337:
The actual cases are:

- $systempkgs = true and $virtualenv_version > 1.7 and $virtualenv_version < 20: 
  $system_pkgs_flag =  "--system-site-packages"

- $systempkgs = false and $virtualenv_version > 1.7 and $virtualenv_version < 20:
  $system_pkgs_flag =  "--no-site-packages"

- $systempkgs = true and $virtualenv_version >= 20: 
  $system_pkgs_flag =  "--system-site-packages"

- $systempkgs = false and $virtualenv_version >= 20:
  $system_pkgs_flag =  ""

so on any virtualenv version we have if $systempkgs = true we can add $system_pkgs_flag =  "--system-site-packages". and based on the virtualenv version we need to figure out what to set for $system_pkgs_flag.

I am not sure if we still need to check on the case if $virtualenv_version < 1.7. On any OSes we are using the virtualenv_version is > 1.7.

PeterJCLaw:
Ah, just seen that https://github.com/voxpupuli/puppet-python/pull/552 was a previous attempt at this.

> The actual cases are:
> 
> * $systempkgs = true and $virtualenv_version > 1.7 and $virtualenv_version < 20:
>   $system_pkgs_flag =  "--system-site-packages"
> 
> * $systempkgs = false and $virtualenv_version > 1.7 and $virtualenv_version < 20:
>   $system_pkgs_flag =  "--no-site-packages"
> 
> * $systempkgs = true and $virtualenv_version >= 20:
>   $system_pkgs_flag =  "--system-site-packages"
> 
> * $systempkgs = false and $virtualenv_version >= 20:
>   $system_pkgs_flag =  ""
>
> so on any virtualenv version we have if $systempkgs = true we can add $system_pkgs_flag = "--system-site-packages". and based on the virtualenv version we need to figure out what to set for $system_pkgs_flag.

From a quick look these seem correct, but incomplete -- there's no handling for `virtualenv` before 1.7. I don't think that `--system-site-packages` exists in the earlier versions.

> I am not sure if we still need to check on the case if $virtualenv_version < 1.7. On any OSes we are using the virtualenv_version is > 1.7.

I would expect that support is required unless the module explicitly declares that it doesn't support the version. (as a user, I'd also hope that it would warn clearly about that somewhere, perhaps at runtime when trying to use an incompatible `virtualenv`).

crazymind1337:
Can you please join #voxpupuli in puppets slack?

crazymind1337:
From `virtualenv --help`:

```
  --no-site-packages    DEPRECATED. Retained only for backward compatibility.
                        Not having access to global site-packages is now the
                        default behavior.
```

crazymind1337:
virtualenv in version 1.7 has been release 9 years ago. I don't find anything if it is already EOL but I think we are safe to drop its support.

https://pypi.org/project/virtualenv/1.7/

I have check all supported OS (from metadata.json). All have virtualenv => 15

crazymind1337:
I think we are safe to just add ```    $system_pkgs_flag = $systempkgs ? {
      true    => '--system-site-packages',
      default => '',
    }```

bastelfreak:
hey people. I created https://github.com/voxpupuli/puppet-python/pull/589 . can you both please take a look? I'm also open for suggestions to properly test if this is working

bastelfreak:
hey people. Since https://github.com/voxpupuli/puppet-python/pull/593 got merged, this issue can be closed?

PeterJCLaw:
Yup, that PR has fixed it for me also.

PR #580, opened 2020-11-17
General linting/indent/syntax

crazymind1337:
I have just adjusted the manifests to current code style. There should not be any compatibility changes.


manifests/gunicorn.pp
```diff
@@ -34,26 +34,26 @@
...
+  String[1]                             $group             = 'www-data',
+  String[1]                             $appmodule         = 'app:app',
+  Variant[Boolean,Hash]                 $osenv             = false,
+  Integer                               $timeout           = 30,
+  Variant[Boolean,Integer]              $workers           = false,
+  Variant[Boolean,String[1]]            $access_log_format = false,
+  Variant[Boolean,Stdlib::Absolutepath] $accesslog         = false,
+  Variant[Boolean,Stdlib::Absolutepath] $errorlog          = false,
+  Python::Loglevel                      $log_level         = 'error',
+  String[1]                             $template          = 'python/gunicorn.erb',
```
igalic, line 55: should we convert this to epp, while we're at it?

manifests/init.pp
```diff
@@ -36,60 +36,48 @@
...
-  Stdlib::Httpurl $anaconda_installer_url         = $python::params::anaconda_installer_url,
-  Stdlib::Absolutepath $anaconda_install_path     = $python::params::anaconda_install_path,
-  Boolean $manage_scl                             = $python::params::manage_scl,
-  Optional[Pattern[/[0-7]{1,4}/]] $umask          = undef,
+  Enum['absent', 'present', 'latest'] $ensure                      = $python::params::ensure,
+  Python::Version                     $version                     = $python::params::version,
+  Enum['absent', 'present', 'latest'] $pip                         = $python::params::pip,
+  Enum['absent', 'present', 'latest'] $dev                         = $python::params::dev,
+  Enum['absent', 'present', 'latest'] $virtualenv                  = $python::params::virtualenv,
+  Enum['absent', 'present', 'latest'] $gunicorn                    = $python::params::gunicorn,
```
igalic, line 44: how is there no stdlib type for these?

manifests/init.pp
```diff
@@ -36,60 +36,48 @@
...
-  }
-
   contain python::install
   contain python::config
 
   Class['python::install']
   -> Class['python::config']
 
-  # Set default umask.
+  ## Set default umask.
```
igalic, line 75: why do we need two `##` here

manifests/init.pp
```diff
@@ -36,60 +36,48 @@
...
   if $umask != undef {
     Exec { umask => $umask }
```
igalic, line 77: doing this seems like bad form

manifests/install.pp
```diff
@@ -247,14 +248,22 @@
         'RedHat': {
           if $pip_ensure != 'absent' {
             if $python::use_epel == true {
-              include 'epel'
-              if $python::manage_pip_package { Class['epel'] -> Package['pip'] }
-              if $python::manage_python_package { Class['epel'] -> Package['python'] }
+              include epel
+
+              if $python::manage_pip_package {
+                Class['epel'] -> Package['pip']
```
igalic, line 254: we could instead just `require epel`, or? cc @alexjfisher

manifests/install.pp
```diff
@@ -247,14 +248,22 @@
...
+              }
+
+              if $python::manage_python_package {
+                Class['epel'] -> Package['python']
```
igalic, line 258: same question here

manifests/install.pp
```diff
@@ -247,14 +248,22 @@
...
+              }
             }
           }
+
           if ($venv_ensure != 'absent') and ($facts['os']['release']['full'] =~ /^6/) {
             if $python::use_epel == true {
-              include 'epel'
+              include epel
+
               Class['epel'] -> Package['virtualenv']
```
igalic, line 267: and here's

manifests/params.pp
```diff
@@ -4,28 +4,32 @@
...
-  $manage_gunicorn           = true
-  $manage_python_package     = true
-  $manage_virtualenv_package = true
-  $manage_pip_package        = true
-  $provider                  = undef
-  $valid_versions            = undef
-  $manage_scl                = true
+  ## Module compatibility check
+  unless $facts['os']['family'] in [ 'Debian', 'RedHat', 'Suse', 'Gentoo', 'AIX', ] {
+    fail("Module is not compatible with ${facts['os']['name']}")
```
igalic, line 9: i wouldn't do this
we have enough hooks into the module to use it on unsupported operating systems, but this makes it impossible

manifests/pyvenv.pp
```diff
@@ -22,15 +22,15 @@
...
-  $environment      = [],
+  $ensure      = present,
+  $version     = 'system',
+  $systempkgs  = false,
+  $venv_dir    = $name,
+  $owner       = 'root',
+  $group       = 'root',
+  $mode        = '0755',
+  $path        = ['/bin', '/usr/bin', '/usr/sbin', '/usr/local/bin'],
+  $environment = [],
```
igalic, line 33: how come we have no types here?

manifests/virtualenv.pp
```diff
@@ -31,29 +31,30 @@
...
+  $group           = 'root',
+  $mode            = '0755',
+  $proxy           = undef,
+  $environment     = [],
+  $path            = ['/bin', '/usr/bin', '/usr/sbin', '/usr/local/bin'],
+  $cwd             = undef,
+  $timeout         = 1800,
+  $pip_args        = '',
+  $extra_pip_args  = '',
+  $virtualenv      = undef,
```
igalic, line 52: this is a public interface, so it should have types

igalic (changes requested):
first pass

manifests/gunicorn.pp
crazymind1337: This would be out of the scope. But @bastelfreak already mentioned this in one of our recent talks @puppetlabs-slack.

manifests/init.pp
crazymind1337: There surely is one. But I didn't want to change too much for now.

manifests/install.pp
bastelfreak: yes

manifests/dotfile.pp
```diff
@@ -24,17 +24,17 @@
 #
 define python::dotfile (
   Enum['absent', 'present'] $ensure   = 'present',
```
bastelfreak, line 26: if you know vim magic to  automatically align it this  way, please let me know :D (I already do `gaip=` to align blocks for `=`).

manifests/gunicorn.pp
bastelfreak: I think epp should one of the followup steps, but we should keep the diff small in each PR.

manifests/dotfile.pp
crazymind1337: Well I do it this manually at Atom.

manifests/init.pp
crazymind1337: Might be more a personal issue, my editor autofixes comments to double #. Since normally comments are double quoted via `//`, `<--` and so on. 

I will remove this :D

manifests/params.pp
crazymind1337: This just has been moved from init.pp to params.pp to have the init.pp code cleaned. I wanted to keep it since I don't wanted to change too much.

manifests/pyvenv.pp
crazymind1337: Ah.. forget this :( sorry.

manifests/virtualenv.pp
crazymind1337: Ah.. forget this :( sorry.

manifests/init.pp
crazymind1337: I have looked up, and didn't find one in stdlib. But I can create a new one with within the module

manifests/init.pp
crazymind1337: Fixed

manifests/install.pp
crazymind1337: Fixed.

manifests/install.pp
crazymind1337: Fixed.

manifests/pyvenv.pp
crazymind1337: Fixed.

manifests/install.pp
crazymind1337: Fixed.

manifests/init.pp
crazymind1337: Why is this a bad form?

manifests/virtualenv.pp
crazymind1337: Fixed.

manifests/init.pp
b4ldr: > Puppet uses dynamic scoping for resource defaults, even though it no longer uses dynamic variable lookup. This means that when you use a resource default statement in a class, it could affect any classes or defined types that class declares. Therefore, they should not be set outside of site.pp. Use per-resource default attributes when possible.

ref: https://puppet.com/docs/puppet/6.19/lang_defaults.html

crazymind1337: Hm okay this looks new to me.

crazymind1337: Fixed.

igalic:
can we open up some follow up issues for the findings, before merging this and forgetting about it forever

manifests/pip/bootstrap.pp
```diff
@@ -40,7 +40,8 @@
         require     => Package['python3'],
         provider    => $exec_provider,
       }
-      # puppet is opinionated about the pip command name
+
+      ## puppet is opinionated about the pip command name
```
igalic, line 44: another `##`

manifests/pip/bootstrap.pp
```diff
@@ -56,7 +57,8 @@
         require     => Package['python'],
         provider    => $exec_provider,
       }
-      # puppet is opinionated about the pip command name
+
+      ## puppet is opinionated about the pip command name
```
igalic, line 61: another `##`

igalic (approved):
minor nit, looks good otherwise

manifests/pip/bootstrap.pp
crazymind1337: Fixed.

manifests/pip/bootstrap.pp
crazymind1337: Fixed.

crazymind1337:
> can we open up some follow up issues for the findings, before merging this and forgetting about it forever

https://github.com/voxpupuli/puppet-python/issues/581

igalic:
added two more issues

manifests/requirements.pp
```diff
@@ -26,21 +26,21 @@
...
+  Stdlib::Absolutepath                         $requirements           = $name,
+  Variant[Enum['system'],Stdlib::Absolutepath] $virtualenv             = 'system',
+  Enum['pip', 'pip3']                          $pip_provider           = 'pip',
+  String[1]                                    $owner                  = 'root',
+  String[1]                                    $group                  = 'root',
+  Optional[Stdlib::HTTPUrl]                    $proxy                  = undef,
+  Any                                          $src                    = false,
+  Array                                        $environment            = [],
+  Boolean                                      $forceupdate            = false,
+  Stdlib::Absolutepath                         $cwd                    = undef,
```
bastelfreak, line 38: this is a backwards-incompatible change. previously you did not need to set a cwd, now it's mandatory. this should be fixed. In the meantime I added the backwards-incompatible label here.

bastelfreak:
fixed it in https://github.com/voxpupuli/puppet-python/pull/587

PR #589, opened 2020-12-05
virtualenv 20x: fix site-packages param

bastelfreak:


bastelfreak:
this was implemented in https://github.com/voxpupuli/puppet-python/pull/593

PR #590, opened 2020-12-07
enhance pyvenv tests

bastelfreak:

### DIFF
--- a/spec/acceptance/pyvenv_spec.rb
+++ b/spec/acceptance/pyvenv_spec.rb
@@ -143,4 +143,46 @@ describe 'python::pyvenv defined resource' do
       apply_manifest(pp, catch_changes: true)
     end
   end
+
+  context 'with versioned minimal python::pip and without systempkgs' do
+    it 'works with no errors' do
+      pp = <<-PUPPET
+      class { 'python':
+        version => '3',
+        dev     => 'present',
+      }
+      user { 'agent':
+        ensure         => 'present',
+        managehome     => true,
+        home           => '/opt/agent',
+      }
+      group { 'agent':
+        ensure => 'present',
+        system => true,
+      }
+      python::pyvenv { '/opt/agent/venv':
+        ensure     => 'present',
+        systempkgs => false,
+        owner      => 'agent',
+        group      => 'agent',
+        mode       => '0755',
+      }
+      python::pip { 'agent' :
+        ensure     => '0.1.2',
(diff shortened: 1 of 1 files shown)


spec/acceptance/pyvenv_spec.rb
```diff
@@ -143,4 +143,44 @@ class { 'python':
       apply_manifest(pp, catch_changes: true)
     end
   end
+  context 'with versioned minimal python::pip and without systempkgs' do
```
ekohl, line 146: I think this hints that rubocop-rspec is not running. There should be a newline separating them.

bastelfreak: updated it

spec/acceptance/pyvenv_spec.rb
```diff
@@ -143,4 +143,45 @@ class { 'python':
...
+        owner      => 'agent',
+        group      => 'agent',
+      }
+      PUPPET
+
+      # Run it twice and test for idempotency
+      apply_manifest(pp, catch_failures: true)
+      apply_manifest(pp, catch_changes: true)
+    end
+    describe command('/opt/agent/venv/bin/pip list') do
```
ekohl, line 182: https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecemptylineafterexample and https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecemptylineafterexamplegroup suggests the same needs to happen here.

bastelfreak: I don't mind adding it, but rubocop doesn't complain at the moment:

```
$ bundle exec rake rubocop:auto_correct
Running RuboCop...
Inspecting 20 files
Parser::Source::Rewriter is deprecated.
Please update your code to use Parser::Source::TreeRewriter instead
....................

20 files inspected, no offenses detected
```

dhollinger: This does make it harder to read. I'd say to add the space and then consider adding that check to modulesync

dhollinger (changes requested):

spec/acceptance/pyvenv_spec.rb
ekohl: > I don't mind adding it, but rubocop doesn't complain at the moment:

I suspect this is because the version is too old to contain it.

bastelfreak:
added it

PR #592, opened 2020-12-12
Fix python::virtualenv to allow virtualenv to not require absolute path

treydock:

I maintain several modules where I might do something like `virtualenv => 'virtualenv-3'` and that works just fine and doesn't require me to know the full path on each OS.  Also the default value if undef is just `virtualenv` so doesn't make sense to require the non-default values be more restrictive than the default.


bastelfreak:
@treydock can you maybe add a test that enforces it's just a string? Otherwise I'm friend this gets changed to Stdlib::Absolutepath soonish again

treydock:
@bastelfreak Added unit tests for virtualenv.

bastelfreak (approved):

PR #593, opened 2020-12-13
Support Ubuntu 20.04

treydock:

Add support for Ubuntu 20.04.  Had to force virtualenv to not use `distribute` otherwise produces errors.  Also updated virtualenv tests to consistently use Python 3 since first set of tests did this but subsequent tests did not and that broke some OS acceptance tests.

Replaces #576

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -284,8 +284,11 @@ class python::install {
         $pip_package  = 'pip'
         $pip_provider = 'pip'
       } else {
+        $pip_package = fact('os.release.major') ? {
+          '20.04' => 'python3-pip',
+          default => 'python-pip',
+        }
         $pip_category = undef
-        $pip_package  = 'python-pip'
         $pip_provider = 'pip'
       }
 
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -284,7 +284,7 @@ class python::install {
         $pip_package  = 'pip'
         $pip_provider = 'pip'
       } else {
-        $pip_package = fact('os.release.major') ? {
+        $pip_package = $facts['os']['release']['major'] ? {
           '20.04' => 'python3-pip',
           default => 'python-pip',
         }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -46,7 +46,7 @@ class python::params {
     default  => 'gunicorn',
   }
 
-  $version = fact('os.release.major') ? {
+  $version = $facts['os']['release']['major'] ? {
     '20.04' => '3',
     default => 'system',
   }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -10,7 +10,6 @@ class python::params {
   }
 
   $ensure                      = 'present'
-  $version                     = 'system'
   $pip                         = 'present'
   $dev                         = 'absent'
   $virtualenv                  = 'absent'
@@ -45,5 +44,10 @@ class python::params {
   $gunicorn_package_name = $facts['os']['family'] ? {
     'RedHat' => 'python-gunicorn',
     default  => 'gunicorn',
+  }
+
+  $version = fact('os.release.major') ? {
+    '20.04' => '3',
+    default => 'system',
   }
 }
--- a/manifests/pyvenv.pp
+++ b/manifests/pyvenv.pp
@@ -48,7 +48,7 @@ define python::pyvenv (
       $python3_venv_package = "python${normalized_python_version}-venv"
 
       case $facts['os']['distro']['codename'] {
-        'xenial', 'bionic', 'cosmic', 'disco', 'stretch', 'buster': {
+        'xenial', 'bionic', 'cosmic', 'disco', 'stretch', 'buster', 'focal': {
           ensure_packages($python3_venv_package)
 
           Package[$python3_venv_package] -> File[$venv_dir]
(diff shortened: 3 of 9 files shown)


bastelfreak (approved):

treydock:
Looks like the Ubuntu 20.02 & Puppet 5 acceptance failure was networking issues boostrapping the beaker nodeset.  Will need to be rerun.

bastelfreak:
Ubuntu 29 + puppet 5 is expected to fail because puppet does not provide packages fro them :(

treydock:
Looks like majority of tests are failing virtualenv with this error:

```
       	Notice: /Stage[main]/Main/Python::Virtualenv[venv]/Exec[python_virtualenv_/opt/venv6]/returns:     AttributeError: module 'importlib._bootstrap' has no attribute 'SourceFileLoader'
```

After much googling at least with Ubuntu 20.04 the online solutions did not work.  The only way around it was not trying to install `distribute` pip package so forcing `distribute => false` for `python::virtualenv`.  I'm wondering if that should become the new default because majority of OSes with Python 3 seem to not work with `distribute => true`.

manifests/pyvenv.pp
```diff
@@ -48,7 +48,7 @@
       $python3_venv_package = "python${normalized_python_version}-venv"
 
       case $facts['os']['distro']['codename'] {
-        'xenial', 'bionic', 'cosmic', 'disco', 'stretch', 'buster': {
+        'xenial', 'bionic', 'cosmic', 'disco', 'stretch', 'buster', 'focal': {
```
bastelfreak, line 51: that's all the ubuntu/debian versions we support. I think we can just drop this case statement?

treydock:
@bastelfreak I am not real familiar with how voxpupuli modulesync sets up `setfiles` for acceptance, my exclude didn't seem to take.   I think the hash used to define matrix causes issues with being able to exclude elements

manifests/pyvenv.pp
treydock: Ya, I dropped it.

bastelfreak:
we havn't figured out yet how to exclude sets :(

bastelfreak (approved):
looks good to me, but will keep it open for a day or two so others can review it as well

bastelfreak:
@crazymind1337 can you take a look as well?

bastelfreak:
@treydock can you rebase against master to get rid of the merge commit?

treydock:
@bastelfreak I had done a merge to get in the virtualenv unit tests which needed to be adjusted for this pull request. Looks to have resolved the conflicts.

manifests/install.pp
```diff
@@ -284,8 +284,11 @@
         $pip_package  = 'pip'
         $pip_provider = 'pip'
       } else {
+        $pip_package = fact('os.release.major') ? {
```
crazymind1337, line 287: Can we use here the $facts-Hash? I know everybody has his favor way to access the facts. But just to be consistent within the module I would like to see the $facts-Hash.

manifests/params.pp
```diff
@@ -46,4 +45,9 @@
     'RedHat' => 'python-gunicorn',
     default  => 'gunicorn',
   }
+
+  $version = fact('os.release.major') ? {
```
crazymind1337, line 49: Same here.

manifests/pyvenv.pp
```diff
@@ -46,15 +46,8 @@
...
-      case $facts['os']['distro']['codename'] {
-        'xenial', 'bionic', 'cosmic', 'disco', 'stretch', 'buster': {
-          ensure_packages($python3_venv_package)
-
-          Package[$python3_venv_package] -> File[$venv_dir]
-        }
-        default: {}
-      }
+      ensure_packages($python3_venv_package)
+      Package[$python3_venv_package] -> File[$venv_dir]
```
crazymind1337, line 50: Could you readd the empty line between `ensure_packages` and the sorting?

manifests/install.pp
treydock: Used fact hash, `$fact['os']['release']['major']`.

manifests/params.pp
treydock: Used fact hash, `$fact['os']['release']['major']`.

manifests/pyvenv.pp
treydock: Added empty line

bastelfreak (approved):

crazymind1337:
While testing on Ubuntu 20.04 I get this result: ```Notice: /Stage[main]/Python::Install/Package[python]/ensure: created
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install python-pip' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package python-pip
Error: /Stage[main]/Python::Install/Package[pip]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install python-pip' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package python-pip```

crazymind1337:
Tested on latest ubuntu:20.04 docker container. Works for me!

crazymind1337 (approved):

crazymind1337:
@treydock You still have to fix the Commit messages. It looks like there is something wrong.

crazymind1337:
> @treydock You still have to fix the Commit messages. It looks like there is something wrong.

Nevermind. I will squash it.

PR #594, opened 2020-12-13
Add Support for RedHat/CentOS 8

treydock:

Ad support for RedHat / CentOS 8

This will cause conflicts with #593 so whichever is merged last will have to be rebased to fix conflicts.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -46,8 +46,11 @@ class python::params {
     default  => 'gunicorn',
   }
 
-  $version = $facts['os']['release']['major'] ? {
-    '20.04' => '3',
-    default => 'system',
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8' {
+    $version = '3'
+  } elsif $facts['os']['release']['major'] == '20.04' {
+    $version = '3'
+  } else {
+    $version = 'system'
   }
 }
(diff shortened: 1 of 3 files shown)


bastelfreak:
I vote for merging https://github.com/voxpupuli/puppet-python/pull/593 first as that has more lines in the diff so rebasing #594 is hopefully easier

bastelfreak (approved):

manifests/params.pp
```diff
@@ -10,7 +10,11 @@
   }
 
   $ensure                      = 'present'
-  $version                     = 'system'
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8' {
+    $version                   = '3'
```
bastelfreak, line 14: feels like we should change `$version` to `'3'` as default for all platforms. We set that in the acceptance tests and don't test anything else. (should be changed in another PR)

treydock: I think the exception should be EL7 because it feels wrong to do something different than what ships with a standard install of the OS.  I think the logic could just be if EL7 , version=2 else version=3.

treydock:
@bastelfreak I've rebased

manifests/params.pp
bastelfreak: I will try to do that in another PR

PR #595, opened 2020-12-15
Set default python to 3

bastelfreak:


bastelfreak:
I was considering setting the default python version on CentOS 7 to python 2. The issue is that we don't test against python 2 since a long time. If I enable python 2 acceptance tests they fail, also python 2 is end of life and centos  7 ships python 3, so it should be used.

kenyon (approved):

bastelfreak:
Ubuntu 20 / Puppet 5 failures are expected because Puppet does not provide packages for that platform combination.

PR #596, opened 2020-12-15
Drop python::virtualenv defined resource

bastelfreak:
The whole virtualenv setup is deprecated upstream. pyvenv / the built-in
python module venv should be used. Since we already have a few backwards
incompatible changes on master, it makes sense to drop this now.


treydock:
@bastelfreak My concern with doing this is it makes it impossible on EL7 to install a virtualenv with system Python.  If I wanted to use system Python 2.7 on EL7 (or EL8) then must use `python::virtualenv` as pyvenv is a Python 3 thing.

bastelfreak:
@treydock yes, but the tests didn't pass on centos 7 with python2 / we  didn't test it since ages.If someone wants to step up and fix it / reimplement it, fine. Otherwise we should drop it (also IMO it's questionable if we should support python 2 because it's really dead).

treydock:
@bastelfreak I'll just be updating my various modules to force Python 3 on EL7.  My main objection is just that RHEL7 and RHEL8 both support Python 2.7 installed from OS repos, so while Python itself might have EOL Python 2.7, RedHat had things downstream of RedHat still support Python 2.7.  Since this module can't manage both Python 3 and Python 2 at same time, just likely gonna have to keep Python 2 management in profiles.

kenyon (approved):

PR #558, opened 2020-08-11
Allow a specific value for the pip version, instead of upgrading virtualenv to latest

GusPriceAtlassian:

PR #599, opened 2021-02-02
Add the possibility to specify the pip version in virtual envs

### LINKED ISSUE #559: Setting Pip Version for Virtual Environments
This is not so much a bug as a feature request. My goal is to set the pip version to a specific version in virtual environments, instead of always creating virtualenvs with "--upgrade pip". I created the following PR to address this:

https://github.com/voxpupuli/puppet-python/pull/558

I'm not sure that this is necessarily the optimal way to get this kind of input, but by adding this variable, when I declare virtualenvs I can now pass a pip version in the format "pip==9.0.3".

SaschaDoering:
Hello,

there is already a pullrequest (https://github.com/voxpupuli/puppet-python/pull/558) that has been opened for https://github.com/voxpupuli/puppet-python/issues/559. But since September 2020 no further progress can be seen. Since this is currently blocking me a bit, I have tried myself on a solution. I hope that's the right way in this case.
I'll add the possibility to set the pip version in virtual envs to bypass a general update to latest. Currently it fails on some of our older Ubuntu installations and can be prevented by specifying a version.

Please let me know if I can support here with some more information.

Thanks

Fixes #559

### DIFF
--- a/manifests/pyvenv.pp
+++ b/manifests/pyvenv.pp
@@ -31,6 +31,7 @@ define python::pyvenv (
   Stdlib::Filemode            $mode        = '0755',
   Array[Stdlib::Absolutepath] $path        = ['/bin', '/usr/bin', '/usr/sbin', '/usr/local/bin',],
   Array                       $environment = [],
+  Python::Venv::PipVersion    $pip_version = 'latest',
 ) {
   include python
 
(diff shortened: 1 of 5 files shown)


SaschaDoering:
I'm not sure, but the CI for Puppet 5 on Ubuntu 20.04 failed probably because of a mistake that has nothing to do with my change.
The two CIs for Ubuntu 16.04 fail with the error that is the reason for my change.
This could be fixed by setting the Pip version. Should I put this in the Acceptance Tests only for Ubuntu 16.04? Or should I rather not fix it so that the tests are not pinned to an old Pip version?

kenyon (dismissed):

kenyon:
> I'm not sure, but the CI for Puppet 5 on Ubuntu 20.04 failed probably because of a mistake that has nothing to do with my change.
> The two CIs for Ubuntu 16.04 fail with the error that is the reason for my change.
> This could be fixed by setting the Pip version. Should I put this in the Acceptance Tests only for Ubuntu 16.04? Or should I rather not fix it so that the tests are not pinned to an old Pip version?

Yes, we do need the tests to pass.

SaschaDoering:
Thanks for your review @kenyon. I fixed the CIs for Ubuntu 16.04 by pinning the Pip version. The tests for Ubuntu 20 & Puppet 5 continues to fail with a network error. This seems to me to be a bug that I can't fix in this module. Or do I oversee something here?

kenyon (approved):

kenyon:
> Thanks for your review @kenyon. I fixed the CIs for Ubuntu 16.04 by pinning the Pip version. The tests for Ubuntu 20 & Puppet 5 continues to fail with a network error. This seems to me to be a bug that I can't fix in this module. Or do I oversee something here?

Yes, Puppet 5 on Ubuntu 20.04 always fails because there are no Puppet 5 packages for Ubuntu 20.04, but we have no way to exclude this combination from the test matrix.

crazymind1337:
@kenyon Can you merge the PR since already have approved it?

crazymind1337:
I will merge this soon since it already has been approved by two persons despite I am from the same company like the requester.

PR #601, opened 2021-03-12
pypa has updated its get-pip.py path

linesight:


kenyon:
The documentation has the existing address, which still works: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py

Where are you seeing that the address has been updated?

Issue #375, opened 2017-06-14
audit metaparameter is deprecated


sfinke0:
Hi,

the following warning occurred on my systems:

```
The `audit` metaparameter is deprecated and will be ignored in a future release.
```

Would you please fix this in an upcoming release?

cheers,
carazzim0

FlorianSW:
The easiest fix seems to be to revert https://github.com/stankevich/puppet-python/commit/7aeaeceda55733767bf357968b3ea4fddef1bd53?

bruhadavid:
Is there an ETA for this?

JvGinkel:
Why not just remove the audit metaparameter from the file resource?

smortex:
This issue is now invalid: the deprecated has been removed and the behavior is still supported:
https://tickets.puppetlabs.com/browse/PUP-8113

This issue can be closed.

PR #612, opened 2021-05-03
Add support for FreeBSD

smortex:

Make this module usable on FreeBSD.

While here, sort operating system lists in alphabetical order for
consistency.

For now, open as a draft so that I can gather early review while I point other modules to this branch and validate it works as expected.  When I am confident with this change, I'll update it.

n/a

### DIFF
--- a/spec/defines/requirements_spec.rb
+++ b/spec/defines/requirements_spec.rb
@@ -53,7 +53,7 @@ describe 'python::requirements', type: :define do
         it { is_expected.to contain_package('gunicorn') }
         it { is_expected.to contain_file('/requirements.txt').with_owner('root').with_group('root') }
 
-        if facts[:os]['name'] == 'Gentoo'
+        if %w[FreeBSD Gentoo].include?(facts[:os]['name'])
           it { is_expected.not_to contain_package('python-dev') }
         else
           it { is_expected.to contain_package('python-dev') }
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -15,11 +15,12 @@ class python::install {
   }
 
   $pythondev = $facts['os']['family'] ? {
-    'AIX'    => "${python}-devel",
-    'RedHat' => "${python}-devel",
-    'Debian' => "${python}-dev",
-    'Suse'   => "${python}-devel",
-    'Gentoo' => undef,
+    'AIX'     => "${python}-devel",
+    'Debian'  => "${python}-dev",
+    'FreeBSD' => undef,
+    'Gentoo'  => undef,
+    'RedHat'  => "${python}-devel",
+    'Suse'    => "${python}-devel",
   }
 
   $pip_ensure = $python::pip ? {
@@ -225,6 +226,10 @@ class python::install {
         $pip_category = undef
         $pip_package  = 'python2-pip'
         $pip_provider = pip2
+      } elsif $facts['os']['family'] == 'FreeBSD' {
+        $pip_category = undef
+        $pip_package  = "py${python::version}-pip"
+        $pip_provider = 'pip'
       } elsif $facts['os']['family'] == 'Gentoo' {
         $pip_category = 'dev-python'
         $pip_package  = 'pip'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,7 +5,7 @@
 #
 class python::params {
   # Module compatibility check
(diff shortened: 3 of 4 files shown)


bastelfreak:
looks quite good. Let's see what CI says

spec/defines/requirements_spec.rb
```diff
@@ -53,7 +53,7 @@
         it { is_expected.to contain_package('gunicorn') }
         it { is_expected.to contain_file('/requirements.txt').with_owner('root').with_group('root') }
 
-        if facts[:os]['name'] == 'Gentoo'
+        if ['FreeBSD', 'Gentoo'].include?(facts[:os]['name'])
```
bastelfreak, line 56: rubocop doesn't like that. Our `rubocop:auto_correct` rake task might fix that automatically for you

smortex: Ah, fixed…

bastelfreak:
@smortex do you want to add more here or is it ready for merge?

smortex:
It is currently not working, but I suspect this is not FreeBSD-related…  I am testing with this simple manifest:

```puppet
class { 'python':
  version => '38',
}

class { 'puppetboard':
}
```

This apply without error, noting happen on subsequent runs, but the requirements from requirements.txt are _not_ installed.  chowning / chmoding the requirements.txt file allows Puppet to trigger the update.  I will have access to a convenient machine for testing tomorrow.  If the problem is not tied to FreeBSD, I think it is mergeable, otherwise more work is indeed needed.

smortex:
I confirm the issue: on Linux the behavior is as expected, but on FreeBSD deployment of the requirements does not work.

I have a working workaround: instead of the trick with managing a resource for the requirements.txt file and auditing changes of it's content, use a python script, executed from the virtualenv which tries to load all dependencies listed in requirements.txt as an `unless` parameter for the `pip install` exec ([The script for checking dependencies](https://stackoverflow.com/a/16298328)).

@bastelfreak do you prefer that we investigate this is a dedicated PR?  If so, this can be merged :wink:

smortex:
Hum, I missed the warning, but it might also help regarding #375 as a side effect :-)

bastelfreak:
where exactly is the issue on FreeBSD? In puppetboard, in python::pyvenv?

smortex:
I am currently investigating it: when trying to reproduce the issue with a minimal case, it is actually working as expected.  I guess it's the way I test that trigger the issue.

Never used `audit` before and there are some surprises :-D

smortex:
@bastelfreak just re-tested from scratch and it works as expected…  I must have screwed something on my other test machine :unamused: 

```sh-session
romain@agrajag ~/Projects % cat > test.pp << EOT
class { 'python':
  version => '38',
}

class { 'puppetboard':
  apache_confd   => '/usr/local/etc/apache24/extra/',
  apache_service => 'apache24',
  python_version => '3.8',
}
EOT
romain@agrajag ~/Projects % git clone https://github.com/voxpupuli/puppet-puppetboard
romain@agrajag ~/Projects/puppet-puppetboard % cd puppet-puppetboard
romain@agrajag ~/Projects/puppet-puppetboard % bundle
romain@agrajag ~/Projects/puppet-puppetboard % bundle exec rake spec_prep
romain@agrajag ~/Projects/puppet-puppetboard % cd spec/fixtures/modules/python
romain@agrajag ~/Projects/puppet-puppetboard/spec/fixtures/modules/python % git remote add smortex https://github.com/smortex/puppet-python
romain@agrajag ~/Projects/puppet-puppetboard/spec/fixtures/modules/python % git fetch smortex
romain@agrajag ~/Projects/puppet-puppetboard/spec/fixtures/modules/python % git checkout freebsd
romain@agrajag ~/Projects/puppet-puppetboard/spec/fixtures/modules/python % cd ../../../..
romain@agrajag ~/Projects/puppet-puppetboard % puppet apply -t ../test.pp --modulepath spec/fixtures/modules
romain@agrajag ~/Projects/puppet-puppetboard % ls /srv/puppetboard/virtenv-puppetboard/lib/python3.8/site-packages
# What is expected \o/
```

This can be merged, thanks!

bastelfreak:
thanks! can you maybe add that as an example to the puppetboard module and add freebsd to the metadata.json in that module?

smortex:
Yeah, it's one of the next move for me, but even more work needed because we do actually do not need the python module to manage puppetboard on FreeBSD because it's as simple as installing a package :wink:.

That being said, this module was ideal for testing these changes quickly ^_^

Thanks!

PR #618, opened 2021-08-01
simplify packages version detection

### LINKED ISSUE #617: ensure not detecting a pip version

this is happening with the latest version of the module: 6.1.0

I am wondering if this is really the intended, and if we need to have a parameter to disable strict version checking:

```
if $_ensure =~ /^((19|20)[0-9][0-9]-(0[1-9]|1[1-2])-([0-2][1-9]|3[0-1])|[0-9]+\.\w+\+?\w*(\.\w+)*)$/ {
```

for instance, I do have different software versions in our environments, as follows: 

```
indico_version: '2.1.11+20210406.1454.2b0d0b9'    # KO
indico_version: '2.3.5+20210720.1912.e31eb12'     # KO
indico_version: '2.0+202106300809.4eec583ae7'     # OK
```
and only the latter is detected. It would accept 2.1.11, but not 2.1.11+xxxxx. 

But rather than trying to fix this convoluted regex, would it make sense to have a parameter called `strict_version_checking` that can be set to true/false?

maxadamo:
Simplify packages version detection

Fixes #617

### DIFF
--- a/spec/acceptance/pip_spec.rb
+++ b/spec/acceptance/pip_spec.rb
@@ -60,7 +60,6 @@ describe 'python::pip defined resource' do
         virtualenv => '/opt/test-venv',
         require => Python::Pip['agent package install'],
       }
-
       PUPPET
 
       apply_manifest(pp, catch_failures: true)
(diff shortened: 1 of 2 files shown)


spec/acceptance/pip_spec.rb
```diff
@@ -71,4 +70,36 @@ class { 'python':
...
+
+      apply_manifest(pp, catch_failures: false)
+      apply_manifest(pp, catch_changes: true)
+    end
+  end
+
+  describe command('/opt/test-venv/bin/pip show agent') do
+    its(:exit_status) { is_expected.to eq 1 }
+    its(:stdout) { is_expected.not_to match %r{WARNING: Package\(s\) not found: agent} }
+  end
```
smortex, line 104: Testing that something does not work in acceptance tests is a bit awkward because you have to test for the very exact error you are expecting.  Here, Puppet cannot apply the catalog (because no package exist with this name and version) so the test should ensure that this is the exact error we are seeing.

It is probably easier to do a unit test in `spec/defines/pip_spec.rb`: setup `python::pip` with your parameters and ensure that the expected resources are declared with the expected parameters.

maxadamo: @smortex I understand, but there is a problem. [Indico](https://indico.cern.ch/) is an application made by CERN, but we are customizing it, and those version numbers come from a colleague of mines. Unfortunately, they're not available in a public repository and they can't be used for this test. 
On another note, this is the first time that I see "Github Actions", and the problem is that it's set to run on PRs, and I can't seem to find a way to run the test on my forked repo. I'll investigate a bit more.

maxadamo: now it's gonna run. It will prove that the package manager is capable of handling a failure on its own.

maxadamo:
I squashed the commits and fixed the CI.

deric:
@maxadamo Any progress on this?

maxadamo:
@deric for me this change is working and it solves a real problem. I try to rebase and solve the conflicts ASAP.

maxadamo:
@deric and @smortex I created a new PR #703 which replaces this one.

PR #619, opened 2021-08-19
If user declares their requirements.txt in Puppet, don't skip pip installation in python::requirements

### LINKED ISSUE #613: Requirements not updated for pyvenv -- requires forceupdate

- Puppet: 6.22.1
- Ruby: ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
- Distribution: CentOS 7.9
- Module version: 6.0.0

```puppet
  file { '/opt/requirements.txt':
    ensure  => present,
    mode    => '0755',
    source  => 'puppet:///modules/venv/requirements.txt',
  }

  python::pyvenv { "${virtual_env}":
    ensure     => present,
    version    => '3.6',
    systempkgs => false,
    venv_dir   => "${virtual_env}",
  }

  python::requirements { '/opt/requirements/requirements.txt':
    virtualenv   => "${virtual_env}",
    pip_provider => 'pip3',
  }
```

## What are you seeing

1. Virtual environment is created.
2. requirements.txt is not applied.

1. Virtual environment is created.
2. requirements.txt is applied.

## Any additional information

* When setting forceupdate to true for python::requirements, the requirements file is applied. But it is applied every puppet run, and aside from this being noisy, with dependencies of packages, this may lead to future problems.
* I think there may be a bug in manifests/requirements.pp:

Now:

```puppet
    $local_subscribe = File[$requirements]
  } else {
    $local_subscribe = undef
  }
```

I think that could be so that $local_subscribe is always set to File[$requirements]:

```puppet
  }
  $local_subscribe = File[$requirements]
```
This fixed the issue for me, but I don't know what the rationale is behind the current behaviour...

acullenn:
@smortex 

Fixes issue wherein the `python::requirements` class does not correctly install a requirements.txt file declared in Puppet to the declared venv, e.g.

```puppet
     file { "${project_dir}/requirements.txt":
        ensure  => present,
        content => file("${module_name}/requirements.txt")
      }
    -> python::pyvenv { "${project_dir}/pyenv":
      ensure  => present,
      version => '3.6'
      }
    -> python::requirements {"${project_dir}/requirements.txt":
        virtualenv => "${project_dir}/pyenv",
      }
```

This appears to be due to the logic for setting the local resource subscription - it checks if the requirements file is `not` declared before setting the local subscription which ultimately gates the pip installation. This behavior runs contrary to the provided Example(s), in which `requirements.txt` must exist or be declared in advance of invoking the class:
https://forge.puppet.com/modules/puppet/python/reference#pythonrequirements
```puppet
python::requirements { '/var/www/project1/requirements.txt' :
  virtualenv => '/var/www/project1',
  proxy      => 'http://proxy.domain.com:3128',
  owner      => 'appuser',
  group      => 'apps',
}
``` 

#### Changes
* Always set the local subscription in `python::requirements`, which is used to trigger the pip installation. 

Fixes #613


smortex:
Ah. I feel like I encountered this at some point, but I was not able to reproduce my failing test after refactoring so assumed I added a missing dependency somewhere…

I would have expected these changes to break tests where multiple venv share the same requirements.txt file but maybe we do not test this case in CI.

Yet, I do not feel this is the right fix since we want to trigger a new install if the requirements file is changed and not managed by Puppet (e.g. it is part of a repo managed by vcsrepo).  I think we now fail here (and it seems not to not have a corresponding test).

Can you add a (failing) unit test that demonstrate the problem you want to fix so that I have a way to reproduce this and think about it?

saz:
@smortex showed a valid reason, why `local_subscribe` might be set to `undefined`. I'm using `vcsrepo` which clones a repo with the `requirements.txt` included, which means, that there's no `File[$requirements]` defined in all cases.

But: this setup won't trigger any installation, as long as `forceupdate` isn't set to true, as `refreshonly` won't help here (there's nothing refreshing anything - I can set this up manually on the `vcsrepo` part).

There's a `--dry-run` option added in pip 22.2 (which is the newest release), which might help here - in the future.
A way to check, if there are uninstalled packages would be the easiest way to resolve this issue, but as far as I know, there's no such thing.

```
python3 -c "import pkg_resources; pkg_resources.require(open('requirements.txt',mode='r'))" &>/dev/null
```
seems to work, but it fails, if there's a mismatch within the package name.
E.g.
```
zope-interface==5.4.0  # requirements.txt
zope.interface==5.4.0  # pip freeze
```
Looks like, as if pip does some magic on the package names (`-` vs `_`, `-` vs `.`,...)

siebrand:
Is there any way to get this patch going again?

zheng460:
Hello, I used Puppet to manage my requirement.txt. This module won't subscribe to the requirement.txt. I have a similar fix that I would like to apply. Is there anything I can do to help get this going?

manifests/install.pp
```diff
@@ -25,10 +25,10 @@
   }
 
   if $python::manage_python_package {
-    package { 'python':
+    ensure_packages (['python'],
```
bastelfreak, line 28: I don't think I like this. There's already a condition around the package resource, so it shouldn't be required to wrap it in `ensure_packages()`

zipkid: This is part of @acullenn 's code changes and i am unsure why this was added.
If they do not reply i will see if i can revert this.

bastelfreak (approved):

PR #616, opened 2021-07-26
New function: any_puppet_to_python()

smortex:
#### Pull Request (PR) description

This function return a String representation of the value passed as
parameter.

The idea is to provide a helper function usable in epp templates when the configuration files we are managing are Python scripts, e.g. https://github.com/voxpupuli/puppet-puppetboard, https://github.com/opus-codium/puppet-taiga

#### This Pull Request (PR) fixes the following issues
n/a

#### Example usage

```epp
EMAIL_USE_TLS = <%= String($taiga::back::email_use_tls, '%T') %>
EMAIL_HOST = <%= if $taiga::back::email_host { String($taiga::back::email_host, '%p') } else { 'None' } %>
```

becomes:

```epp
EMAIL_USE_TLS = <%= any_puppet_to_python($taiga::back::email_use_tls) %>
EMAIL_HOST = <%= any_puppet_to_python($taiga::back::email_host) %>
```

PR #620, opened 2021-08-23
Delete the any_puppet_to_python() function

smortex:
It was integrated into the stdlib as `to_python()`:

- https://github.com/puppetlabs/puppetlabs-stdlib/pull/1200

Label this PR as _skip-changelog_ since no release of the module ship with this function.  #616 which introduced the function has also been labeled _skip-changelog_.


ghoneycutt:
This has not been released yet into a tagged version of stdlib. When it is, this PR will need to update the `metadata.json` to show the minimum requirement.

smortex:
Technically, this module does not need/use/rely on this function, that's why this removal of a function is not a _backward-incompatible_ change.

Modules which want to use this function will indeed have to update the minimum version of stdlib accordingly when a release is made (just as they would have had to update the minimum version of the python module if the function was release in the python module).

ghoneycutt:
Ah right! Suggest adding a note in the README that these functions are removed and you need version X of stdlib.

PR #627, opened 2021-11-14
Remove duplicate arguments in the pip install command

zanyou:

`${install_args}` is duplicated in the arguments given to the `pip install` command to be executed.
The duplication does not cause anything bad, but I think it is better to remove it.

None


kenyon:
I think it would be good to add `install_args` to one of the tests.

zanyou:
@kenyon I've added a new test.

kenyon (approved):

PR #628, opened 2021-11-15
RHEL8 has a different package name

yakatz:
There is no package in EPEL for RHEL8 with the name `python-gunicorn`, but there is `python3-gunicorn`

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -42,7 +42,10 @@ class python::params {
   }
 
   $gunicorn_package_name = $facts['os']['family'] ? {
-    'RedHat' => 'python-gunicorn',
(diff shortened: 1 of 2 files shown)


kenyon:
Thanks for the contribution. Would be great to add a test to https://github.com/voxpupuli/puppet-python/blob/master/spec/classes/python_spec.rb for the RHEL section, similar to the gunicorn tests for the other distributions.

yakatz:
@kenyon Tests added

kenyon (approved):

yakatz:
@bastelfreak Any idea when this might make a release?

bastelfreak:
@yakatz I need to get https://github.com/voxpupuli/puppet-python/pull/631 merged before a release

PR #631, opened 2021-12-10
modulesync 5.1.0 & puppet-lint: fix params_empty_string_assignment

bastelfreak:


TuningYourCode (approved):

bastelfreak:
IMO the switch from `''` to `undef` for the parameters isn't a breaking change. But I didn't add the `modulesync` label on purpose, so this shows up in the changelog.

mmerfort (approved):

PR #633, opened 2022-04-04
cleanup references to obsolete virtualenv parameter/command

vchepkov:
remove references to obsolete virtualenv parameter and obsolete fact


metadata.json
```diff
@@ -7,11 +7,9 @@
   "source": "https://github.com/voxpupuli/puppet-python.git",
   "project_page": "https://github.com/voxpupuli/puppet-python",
   "issues_url": "https://github.com/voxpupuli/puppet-python/issues",
-  "description": "Puppet module for Python",
```
kenyon, line 10: Why should the description be removed?

vchepkov: because it's an obsolete attribute, was deprecated long time ago

smortex (approved):
This makes sense I think.  The relevant code was removed in #596.

metadata.json
smortex: `summary` is terser that this `description` we remove.  Maybe we can update the summary with this text or rephrase it to indicate the module capabilities?

The module description on GitHub says *Puppet module for installing and managing Python, pip, virtualenvs and Gunicorn virtual hosts.* as the README.  Maybe we can adjust this and sync the sentences?

vchepkov: I can remove the change from this PR if you like. I removed the line, because IDE told me it's wrong and I was touching the file anyway. summary suppose to be terse, imho, documentation says "one liner"

vchepkov: rebased/changed summary

smortex: That's great for me, thanks!

kenyon (approved):

PR #641, opened 2022-09-01
Add prompt parameter to python::pyvenv

ookisan:

Add a prompt parameter to python::pyvenv that uses the --prompt command-line argument available in the venv module in Python 3.6 and later, to set the prompt shown when the virtualenv is active.

No issues.

### DIFF
--- a/manifests/pyvenv.pp
+++ b/manifests/pyvenv.pp
@@ -10,6 +10,7 @@
 # @param mode  Optionally specify directory mode
 # @param path Specifies the PATH variable.
 # @param environment Optionally specify environment variables for pyvenv
+# @param prompt Optionally specify the virtualenv prompt (python >= 3.6)
 #
 # @example
 #   python::pyvenv { '/var/www/project1' :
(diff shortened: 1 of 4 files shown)


manifests/pyvenv.pp
```diff
@@ -31,6 +32,7 @@
   Stdlib::Filemode            $mode        = '0755',
   Array[Stdlib::Absolutepath] $path        = ['/bin', '/usr/bin', '/usr/sbin', '/usr/local/bin',],
   Array                       $environment = [],
+  Variant[Boolean,String[1]]  $prompt      = false,
```
smortex, line 35: I feel like setting it to `true` does not make sense, i would rather:

```suggestion
  Optional[String[1]]  $prompt             = undef,
```

ookisan: I don't disagree. I just used Variant because it was the most common variant in existing code. I'll change and use the opportunity to add a test case that ensures --prompt is not used on python 3.5.

bastelfreak (approved):

ookisan:
The failing tests seem to have nothing to do with the code:

`2022-09-02T10:26:59.5082100Z [00;00m[00;31m{"message":"failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: \"/sbin/init\": stat /sbin/init: no such file or directory: unknown"}`

I get the same locally, didn't yesterday. Will this prevent merging and if so, is there anything I can do about it?

bastelfreak:
Ubuntu updated their docker image a few hours ago and I'm wondering if it's broken :thinking:

smortex (approved):

PR #642, opened 2022-09-27
Implement Arch Linux support

bastelfreak:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -38,7 +38,7 @@
 #
 class python (
   Python::Package::Ensure    $ensure                      = $python::params::ensure,
-  Python::Version            $version                     = '3',
+  Python::Version            $version                     = $facts['os']['family'] ? { 'Archlinux' => 'system', default => '3' },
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -38,7 +38,7 @@
 #
 class python (
   Python::Package::Ensure    $ensure                      = $python::params::ensure,
-  Python::Version            $version                     = '3',
+  Python::Version            $version                     = $facts['os']['family'] ? { 'Archlinux' => 'system', default => '3' },
```
evgeni, line 41: Any reason not to put that as `$python::params::version`?

The result should be the same, but I kinda find it odd to have most OS-dependant things defined in `params` but not this one?

bastelfreak: I put it here on purpose because puppet-strings will put it into reference.md and people will see the actual value. if it's in params.pp, it would just say `$python::params::version` and you need to look it up by yourself.

evgeni (approved):

PR #643, opened 2022-09-27
use ensure_packages to install python

MephistoaD:

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -24,10 +24,7 @@ class python::install {
   }
 
   if $python::manage_python_package {
-    package { 'python':
-      ensure => $python::ensure,
-      name   => $python,
-    }
+    ensure_packages('python')
   }
 
   if $python::manage_venv_package {


manifests/install.pp
```diff
@@ -24,10 +24,7 @@
   }
 
   if $python::manage_python_package {
-    package { 'python':
-      ensure => $python::ensure,
-      name   => $python,
-    }
+    ensure_packages('python')
```
smortex, line 27: The actual package name is in `$python`.  I guess this also break any `$python::ensure` which is not `present`…

smortex (commented):
This usage of `ensure_packages()` seems an anti-pattern. What use case are you trying to fix? Doesn't `$python::manage_python_package` allow you to do what you want?

manifests/install.pp
Haribo112: It's just better in general to use `ensure_packages()` to avoid duplicate declaration of resources. According to the [changelog](https://github.com/voxpupuli/puppet-python/blob/master/CHANGELOG.md#1710-2014-09-25) this was already added in version 1.7.10 but according to the installation manifests it is not actually implemented.

PR #651, opened 2023-03-03
Fix pip installation on Gentoo

jameslikeslinux:
The 'pip' package name is ambiguous on Gentoo, with options 'app-misc/pip' and 'dev-python/pip', resulting in the following error during installation:

> Error: /Stage[main]/Python::Install/Package[pip]: Could not evaluate: More than one package with the specified name [pip], please use the category parameter to disambiguate

The error message directs us to use the `category` parameter, which this module already did; however, `category` is a "read-only parameter set by the package" [1].

The correct way to disambiguate this package is to use the full package name, 'dev-python/pip'.

[1] https://www.puppet.com/docs/puppet/7/types/package.html#package-attribute-category

n/a

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -213,42 +213,33 @@ class python::install {
       }
 
       if String($python::version) =~ /^python3/ {
-        $pip_category = undef
         $pip_package  = "${python}-pip"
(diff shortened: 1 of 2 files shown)


smortex (approved):
LGTM

kenyon (approved):

PR #652, opened 2023-03-21
Prefer Hiera data to params.pp

smortex:
This limit the quantity of spaghetti code we add when adding support for
another platform.

No functional change.


data/common.yaml
```diff
@@ -0,0 +1,11 @@
+---
+python::version: '3'
+python::use_epel: true
+python::gunicorn_package_name: 'gunicorn'
+python::manage_pip_package: true
+python::manage_venv_package: true
+python::pip::group: 'root'
```
smortex, line 7: Kudos to who can tell me what is wrong here: the default value is not catch by Automatic Parameter Lookup in `spec/defines/pip_spec.rb`.  Setting an explicit value in the test suite or a default value in `manifests/pip.pp` for the `$group` parameter workaround the issue, but I don't get why it is not working as expected.

smortex: Also, setting values for sub-classes of python is unusual (and maybe not supported?) but I would like to avoid breaking backwards compatibility in this PR.

data/common.yaml
```diff
@@ -0,0 +1,11 @@
+---
+python::version: '3'
```
bastelfreak, line 2: can you please keep the data in the init.pp instead common.yaml? That enables puppet-strings to pick it up and document it in the REFERENCE.md.

bastelfreak: at https://voxpupuli.org/docs/reviewing_pr/, we have:
> Are hiera yaml files added for data-in-modules? Ensure that the data is compatible with [hiera 5](https://puppet.com/docs/puppet/5.3/hiera_migrate.html#use-cases-for-upgrading-to-hiera-5). Static data that is equal across every supported operating system must stay in the init.pp, it shouldn’t be moved to a common.yaml due to [puppet-strings issue #250](https://github.com/puppetlabs/puppet-strings/issues/250).

smortex: These parameters in `common.yaml` have overrides in other `data/**/*.yaml` and should be os-dependent.  They where previously documented as:

```
Default value: `$python::params::<foo>`
```

this line is not shown anymore.  If we put the value in `init.pp` instead of `common.yaml`, wrong default values would be documented for some operating systems.

Issue #668, opened 2023-09-15
Rework package management to allow composition of dependencies


smortex:
This issue has the purpose of tracking progress for improving the way the puppet-python module manage packages.

## Issues we want to solve

In these example, we assume a control-repo with a `profile::python` class that wraps the puppet-python module with site-specific config:

```puppet
class profile::python {
  class { 'python':
    ensure  => present,
    # ...
  }
}
```

### System packages with dependencies on `python-foo`

The `syslog-ng-mod-python` package depend on `python-venv`.  When this package is added to the catalog, puppet will install `syslog-ng-mod-python`, which will install `python-venv` as a dependency.  On next run, `python-venv` is removed (because the python class default to ensure this package is absent) and `syslog-ng-mod-python` which depend on it is also removed.

To workaround this issue, we can skip management of the `python-venv` in the python module with:

```puppet
class profile::python {
  class { 'python':
    # ...
    manage_venv_package => false, # python-venv is a dependency of syslog-ng-mod-python
  }
}
```

If another node need to have the venv package installed, we have a few solutions:
1. Add a parameter to the `profile::python` and use Hiera to set it to a custom value on the nodes that must have different behavior;
2. Duplicate and adjust the `profile::python` class (e.g. `profile::python_with_venv`) and include one or the other of these two conflicting profiles, leading to some duplication;
3. Manage the `python-venv` package in the profile that need it, duplicating what the puppet-python module is supposed to manage.

None of these solutions is great.

### Puppet modules needing a specific component

The puppet-puppetboard module rely on the puppet-python module.  When installing from a venv, it include [this code](https://github.com/voxpupuli/puppet-puppetboard/blob/c23187c7bb9b7276c1f5005b11ee5bd0112ad097/manifests/init.pp#L268-L272) in the catalog:

```puppet
    class { 'python':
      version => $python_version,
      dev     => 'present',
      venv    => 'present',
    }
```

This clash with the above `profile::python`.  To cope with this, the module provide a `manage_virtualenv` parameter to skips this fragment and allow the user to handle this themself, with the same solutions and same drawbacks as seen in the previous example.

## How puppetlabs-apache handle a similar issue

Apache has a lot of different modules, but each one is managed in a different class allowing to do some composition of the dependencies from another puppet module, e.g.:

```puppet
class taiga::vhost {
  include apache
  include apache::mod::passenger

  # ...
}
```

[...]

PR #670, opened 2023-09-15
Split package installation in multiple classes

smortex:
Install python packages form dedicated classes of the module.

This allow to use a single python profile in a control-repo that does not enforce installation / removal of these packages and still allow modules to require them.

```puppet
class foo {
  include python
  include python::install::dev

  # ...
}

class bar {
  include python
  include python::install::venv

  # ...
}


class profile::python {
  class { 'python':
    dev                 => 'present',
    pip                 => 'present',
    venv                => 'present',
    manage_dev_package  => false,
    manage_pip_package  => false,
    manage_venv_package => false,
  }
}

class profile::foo {
  include profile::python

  class { 'foo':
  }
}

class profile::bar {
  include profile::python

  class { 'bar':
  }
}
```

If a node is classified `profile::foo`, the `python-dev` package will be installed.  If a node is classified `profile::bar`, the `python-venv` package will be installed. Other `python-*` packages are *not* managed.

This is part 2/3 of #668 and does not introduce functional change.  However it permits new usages and therefore is labelled enhancement.

Part 3/3 will be backwards-incompatible in order to change the defaults of the `python` class to match those in the above example.


manifests/install.pp
```diff
@@ -32,32 +32,17 @@
...
-    ## CentOS has no extra package for venv
-    ##
-    unless $facts['os']['name'] == 'CentOS' {
-      package { 'python-venv':
-        ensure  => $python::venv,
-        name    => "${python}-venv",
-        require => Package['python'],
-      }
-    }
+    include python::install::venv
```
bastelfreak, line 35: should we use contain here?

PR #675, opened 2023-11-23
Remove legacy top-scope syntax

smortex:


smortex:
PR title adjusted, will show up in future releases.

PR #677, opened 2023-11-27
Add OracleLinux/AlmaLinux/Rocky support

JakeTRogers:

Newer(8+) RedHat family OSes do not package python-venv as it is built in to the versions of Python they ship

n/a

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -35,7 +35,7 @@ class python::install {
     ##
     ## CentOS has no extra package for venv
     ##
-    unless $facts['os']['name'] == 'CentOS' {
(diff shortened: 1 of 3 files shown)


kenyon (approved):
Can you make a separate PR for the change to `metadata.json` so that those changes appear in our changelog correctly? Thanks. Otherwise LGTM.

bastelfreak:
@kenyon I would just rename it to 'Add OracleLinux/AlmaLinux/Rocky support'. Are you fine with that? I don't consider this a bugfix becaue we never claimed to support those in the first place.

kenyon:
@bastelfreak good idea, I updated the PR title.

PR #682, opened 2024-01-01
Make latest version detection portable

smortex:
This is currently broken on FreeBSD do to the usage of GNU-grep specific
grep(1) flag: `-P`.

We can use a single sed(1) command with an extended regular expression
to achieve the same goal, so use this instead.

### DIFF
--- a/manifests/pip.pp
+++ b/manifests/pip.pp
@@ -220,7 +220,7 @@ define python::pip (
         $latest_version = join([
             "${pip_install} ${legacy_resolver} ${pypi_index} ${pypi_extra_index} ${proxy_flag}",
             " ${install_args} ${install_editable} ${real_pkgname}==notreallyaversion 2>&1",
-            ' | grep -oP "\(from versions: .*\)" | sed -E "s/\(from versions: (.*?, )*(.*)\)/\2/g"',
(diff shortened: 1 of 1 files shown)


manifests/pip.pp
```diff
@@ -220,7 +220,7 @@
         $latest_version = join([
             "${pip_install} ${legacy_resolver} ${pypi_index} ${pypi_extra_index} ${proxy_flag}",
             " ${install_args} ${install_editable} ${real_pkgname}==notreallyaversion 2>&1",
-            ' | grep -oP "\(from versions: .*\)" | sed -E "s/\(from versions: (.*?, )*(.*)\)/\2/g"',
+            " | sed -nE 's/.*\\(from versions: (.*, )*(.*)\\)/\\2/p'",
```
smortex, line 223: 1. `-n` prevent output by default (this allows us to skip grep(1) which was used to only process the line matching 'from versions';
2. The removed `?` repetition char (0 or 1) was bogus after the `*` repetition char (0 or any number), so drop it;
3. the `g` flag is used to replace each occurrences, but we can only have a single match on a line, to remove it and add the `p` flag to print the result of the substitution which is not done automatically because of 1. above.

kenyon (approved):

PR #686, opened 2024-02-08
Add python_path to pyvenv class

### LINKED ISSUE #685: Not possible to use multiple python versions for virtualenvs
Currently, there seems to be no way to configure a custom python version installed by something like pyenv. 
Using pyenv, you can install multiple python versions side-by-side without overwriting the system python, yet maintain application specific python versions. 

The linked PR will solve this by utilizing a variable in `python::pyvenv` called `python_path`. If set, it will override the python version used to create that specific virtual environment.

wmellema:
Issue #685

- Add: `python_path` variable to `python::pyvenv`

Add a custom path for python to the pyvenv module in order to use `pyenv` managed python installations

Fixes #685

### DIFF
--- a/manifests/pyvenv.pp
+++ b/manifests/pyvenv.pp
@@ -11,6 +11,7 @@
 # @param path Specifies the PATH variable.
 # @param environment Optionally specify environment variables for pyvenv
 # @param prompt Optionally specify the virtualenv prompt (python >= 3.6)
+# @param python_path Optionally specify python path for creation of virtualenv
 #
 # @example
 #   python::pyvenv { '/var/www/project1' :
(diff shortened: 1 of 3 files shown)


manifests/pyvenv.pp
```diff
@@ -34,6 +35,7 @@
   Array                       $environment = [],
   Optional[String[1]]         $prompt      = undef,
   Python::Venv::PipVersion    $pip_version = 'latest',
+  Optional[String[1]]         $python_path = undef,
```
bastelfreak, line 38: Should this be ?
```suggestion
  Optional[Stdlib::Absolutepath] $python_path = undef,
```

wmellema:
@bastelfreak You are absolutely correct. Added to the PR and updated the documentation

PR #688, opened 2024-02-13
Correct FreeBSD Python package name 

rtprio:

When trying to install Puppetboard I ran into this module trying to install virtualenv (doesn't exist as a package, only a port) and `py3.9-pip` (correct name is `py39-pip`). 

Replace out the `.` in the version name. I did not test this extensively, just enough to get it to compile and install the right package for puppetboard. The origination of of the module doesn't seem to allow an easy way to get a default version based on OS and pass it down to python, pip, etc.

Feedback welcome.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -222,7 +222,7 @@ class python::install {
         $pip_package  = 'python2-pip'
         $pip_provider = pip2
       } elsif $facts['os']['family'] == 'FreeBSD' {
-        $pip_package  = "py${python::version}-pip"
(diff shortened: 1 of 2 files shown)


smortex:
Regarding the version, I never hit that issue.  It happen I use this in my python profile:

```
class profile::python {
  class { 'python':
    version => '39',
    dev     => 'present',
  }
}
```

The data type for the version parameter accept a large number of formats, maybe too much.


The other change seems unrelated and break CI.  Maybe worth a different PR?

PR #693, opened 2024-06-17
Fix pre-release version resolution

### LINKED ISSUE #690: Puppet python latest version resolving - Prereleases
Hello,

There is an issue during the resolution of the latest version from the pip repository:

It takes the last version that is retrieved by the pip. However, it might not be the latest version but a prerelease/development version.

```
" | sed -nE 's/.*\\(from versions: (.*, )*(.*)\\)/\\2/p'",
```

Example of the output for a private repository with a prerelease version.

```pip3 --log /tmp/pip.log install --use-deprecated=legacy-resolver    pydtmux==notreallyaversion 2>&1 | grep -oP "\(from versions: .*\)"
(from versions: 1.0.0, 1.0.1, 1.0.3, 1.0.4, 1.0.6, 1.0.7, 1.0.8, 1.0.9, 1.0.10, 1.1.0, 1.1.1, 1.1.2, 1.1.3, 1.2.0, 1.3.0, 1.3.1, 1.3.2, 1.3.3, 1.3.4, 1.4.0, 1.4.1, 1.5.0, 1.6.0, 1.6.1, 1.7.0, 1.8.0, 1.9.0, 1.9.1, 1.9.2, 1.10.0, 1.10.1, 1.10.2, 1.10.3, 1.10.4, 1.10.5, 2.0.0rc1, 2.0.0rc2, 2.0.0rc3)
```

```
--pre
Include pre-release and development versions. By default, pip only finds stable versions.
```

teluq-pbrideau:
Fix the resolution of the latest version from the pip repository, by removing the pre-release version from the list of the latest package available.  See #690 for details.

#### NOTE
With this fix, if the pip package is not a stable release, it will need an explicit `--pre`:
```
python::pip { 'foo' :
  install_args => '--pre',
}
```

Another option could be to add a new parameter `pre => true`, but I chose to use the `install_args` in the end for this PR, which is more flexible.

Fixes #690


teluq-pbrideau:
Well, thinking more about it, it is probably better for me to parse `install_args` for the `--pre` flag, because it does not create a breaking change…  I’ll modify the PR.

bastelfreak:
Hi @teluq-pbrideau , we recently merged https://github.com/voxpupuli/puppet-python/pull/696 . can you please rebase?

manifests/pip.pp
```diff
@@ -215,14 +215,21 @@
...
+              ' | tr -d "[:space:]"',
+          ])
+        } else {
+          $latest_version = join([
+              "${pip_install} ${legacy_resolver} ${pypi_index} ${pypi_extra_index} ${proxy_flag}",
+              " ${install_args} ${install_editable} ${real_pkgname}==9!0dev0+x 2>&1",
+              " | sed -nE 's/.*\\(from versions: ([^\\)]*)\\)/\\1/p'",
+              ' | awk \'BEGIN {RS=", "} {if ($0 !~ /(a|b|rc|dev)/) {gsub(/\n/,"");stable[arraylen++]=$0}} END {print stable[arraylen-1] }\'',
+          ])
+        }
```
kenyon, line 232: It would be better to avoid duplicating the first two lines of the command, like this:

```puppet
        $version_processing_commands = if $install_args and $install_args =~ /--pre/ {
          [
            " | sed -nE 's/.*\\(from versions: (.*, )*(.*)\\)/\\2/p'",
            ' | tr -d "[:space:]"',
          ]
        } else {
          [
            " | sed -nE 's/.*\\(from versions: ([^\\)]*)\\)/\\1/p'",
            ' | awk \'BEGIN {RS=", "} {if ($0 !~ /(a|b|rc|dev)/) {gsub(/\n/,"");stable[arraylen++]=$0}} END {print stable[arraylen-1] }\'',
          ]
        }

        $latest_version = join(
          [
            "${pip_install} ${legacy_resolver} ${pypi_index} ${pypi_extra_index} ${proxy_flag}",
            " ${install_args} ${install_editable} '${real_pkgname}==9!0dev0+x' 2>&1",
          ] + $version_processing_commands,
        )
```

I think we want to keep the single quotation marks around `${real_pkgname}==9!0dev0+x` too, right?

kenyon (changes requested):

teluq-pbrideau:
I’ve discovered a potential difference when a package have only pre-release version.  I’ve tried with the first package i’ve found with this criteria: `glint4py`: https://pypi.org/project/glint4py/#description

When package installed with `pip install`, the package is installed, but with the `python::pip` class with my change, it is required to add the `install_args => '--pre'` parameter.  Not sure it is desired…

Also, I cannot find a package with pre-release and a previous stable release to test as I did with pip==24.1.b2

kenyon (approved):

teluq-pbrideau:
@kenyon just to make sure you understood my previous comment:
Every current user of this module will have a breaking change and will have to add `install_args => '--pre'` if there is no stable release on pypi

kenyon:
@teluq-pbrideau thanks for the clarification. I'm not using this functionality of this module, so this doesn't really affect me, but it seems like it's OK if you have to explicitly do something to enable installation of prereleases, even if `pip` itself doesn't behave this way. But maybe we wait for more input from users to see whether the breaking change is acceptable.

I know some tools like `pipenv` have extra difficulty dealing with prereleases, so this problem has some precedent in the Python world. The popular formatter `black` caused a lot of consternation for years because it was only a prerelease: https://github.com/psf/black/issues/517

teluq-pbrideau:
@kenyon I rebase with `master`, I believe the failure are not related to my changes...
> Couldn't find id: {"stream":"STEP 1/10: FROM amd64/debian:11\n"}
> Couldn't find id: {"stream":"STEP 1/9: FROM amd64/almalinux:9\n"}

PR #699, opened 2024-08-21
Add option to install python-setuptools

bastelfreak:
This is required at least on Arch Linux because distutils is provided by python-setuptools, not python:

https://gitlab.archlinux.org/archlinux/packaging/packages/thefuck/-/issues/1


sebastianrakel (approved):

manifests/init.pp
```diff
@@ -61,6 +62,7 @@
   Stdlib::Absolutepath       $anaconda_install_path       = '/opt/python',
   Boolean                    $manage_scl                  = true,
   Optional[Python::Umask]    $umask                       = undef,
+  Boolean $manage_setuptools = $facts['os']['family'] ? { 'Archlinux' => true, default => false, },
```
kenyon, line 65: Could also just do

```suggestion
  Boolean $manage_setuptools = $facts['os']['family'] == 'Archlinux',
```

spec/classes/python_spec.rb
```diff
@@ -19,8 +19,10 @@
 
         if facts[:os]['family'] == 'Archlinux'
           it { is_expected.not_to contain_package('pip') }
+          it { is_expected.to contain_package('python-setuptools') }
```
kenyon, line 22:
```suggestion
          it { is_expected.to contain_package('python-distutils-extra') }
```

spec/classes/python_spec.rb
```diff
@@ -19,8 +19,10 @@
...
         else
           it { is_expected.to contain_package('pip') }
+          it { is_expected.not_to contain_package('python-setuptools') }
```
kenyon, line 25:
```suggestion
          it { is_expected.not_to contain_package('python-distutils-extra') }
```

PR #701, opened 2024-09-13
modulesync 9.3.0

bastelfreak:
modulesync 9.2.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@
 
 name: CI
 
+# yamllint disable-line rule:truthy
 on:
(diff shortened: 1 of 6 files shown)


jhoblitt:
@bastelfreak Paging the Arch Lord.

bastelfreak:
>   Error: invalid byte sequence in US-ASCII
  Error: /Stage[main]/Main/Python::Pyvenv[/opt/test-venv]/Exec[python_virtualenv_/opt/test-venv]/returns: change from 'notrun' to ['0'] failed: invalid byte sequence in US-ASCII

something in that exec resource isn't returning utf8 :thinking:

PR #716, opened 2025-04-24
Fix bootstrapping `python::pyvenv` when Python is not installed

smortex:
When python is not installed, a catalog with a `python::pyvenv` resource should compile and apply successfully.

Instead it currently cause a catalog compilation error:
```
   Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, The function 'split' was called with arguments it does not accept. It expects one of:
    (String str, String pattern)
      rejected: parameter 'str' expects a String value, got Undef
    (String str, Regexp pattern)
      rejected: parameter 'str' expects a String value, got Undef
    (String str, Type[Regexp] pattern)
      rejected: parameter 'str' expects a String value, got Undef
    (Sensitive[String] sensitive, String pattern)
      rejected: parameter 'sensitive' expects a Sensitive[String] value, got Undef
    (Sensitive[String] sensitive, Regexp pattern)
      rejected: parameter 'sensitive' expects a Sensitive[String] value, got Undef
    (Sensitive[String] sensitive, Type[Regexp] pattern)
      rejected: parameter 'sensitive' expects a Sensitive[String] value, got Undef (file: /etc/puppetlabs/code/environments/production/modules/python/manifests/pyvenv.pp, line: 48, column: 34) (file: /etc/puppetlabs/code/environments/production/modules/taiga/manifests/back/install.pp, line: 15) on node debian12-64-puppet7.example.com
```

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -1,0 +1,1 @@
+python::default_system_version: "3.11"
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -67,6 +67,7 @@ class { 'python' :
 
 The following parameters are available in the `python` class:
 
+* [`default_system_version`](#-python--default_system_version)
 * [`ensure`](#-python--ensure)
 * [`version`](#-python--version)
 * [`pip`](#-python--pip)
@@ -91,6 +92,12 @@ The following parameters are available in the `python` class:
 * [`rhscl_use_public_repository`](#-python--rhscl_use_public_repository)
 * [`anaconda_installer_url`](#-python--anaconda_installer_url)
 * [`anaconda_install_path`](#-python--anaconda_install_path)
+
+##### <a name="-python--default_system_version"></a>`default_system_version`
+
+Data type: `Python::Version`
+
+The default version of Python provided by the operating system. Only used as a fallback if Python is not installed yet to determine how to handle some actions that vary depending on the Python version used.
 
 ##### <a name="-python--ensure"></a>`ensure`
 
--- a/data/os/Archlinux.yaml
+++ b/data/os/Archlinux.yaml
@@ -1,0 +1,1 @@
+python::default_system_version: "3.13"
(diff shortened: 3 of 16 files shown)


data/common.yaml
```diff
@@ -0,0 +1 @@
+python::default_system_version: "3.11"
```
smortex, line 1: For the common file, the choice of `3.11` was arbitrary. We really only want a default for systems missing in Hiera data (e.g. operating systems not listed as supported).  For example the test suite mocks Debian 6, and we do not want to add support for this in Hiera :upside_down_face:

smortex:
This fix the issue I am facing:

https://github.com/opus-codium/puppet-taiga/actions/runs/14652911906?pr=78

Ready for review!

kenyon (approved):
Nice, this reminds me of what I was talking about in https://github.com/voxpupuli/puppet-puppetboard/pull/403. We have essentially duplicated your system-version work here in the puppetboard hiera data: https://github.com/voxpupuli/puppet-puppetboard/tree/88d2e9a038ae9da14344d422397f6708df2ba9b4/data/os

smortex:
> Nice, this reminds me of what I was talking about in [voxpupuli/puppet-puppetboard#403](https://github.com/voxpupuli/puppet-puppetboard/pull/403). We have essentially duplicated your system-version work here in the puppetboard hiera data: [voxpupuli/puppet-puppetboard@`88d2e9a`/data/os](https://github.com/voxpupuli/puppet-puppetboard/tree/88d2e9a038ae9da14344d422397f6708df2ba9b4/data/os)

Interesting.

1. It allows to compare versions and I only see a difference for RedHat 8 (3.6 vs 3.8).  I found 3.6 [here](https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8-3#) but maybe it was updated in later 8.x RedHat versions? It looks like 3.6 is the last version where the module behavior change, so the impact should be negligible I think :shrug:.
2. If I understand correctly, this should allow to simplify the puppet-puppetboard module when this is merged, right?

smortex:
Fine, let's merge this, make a new release and allow to simplify consumer ports!

PR #733, opened 2025-09-23
Allow setting `python::version` in hiera to `Integer`

### LINKED ISSUE #731: Define pip_package in hieradata
On FreeBSD the Python package name can vary.
The package name is defined in[ manifests/install.pp](https://github.com/voxpupuli/puppet-python/blob/master/manifests/install.pp): `$pip_package  = "py${python::version}-pip"`
I'm looking for a way to define this in hieradata, with that we can easily define for each server what the package name is.

An example error message I am seeing on one of my FreeBSD boxes:

```
Execution of '/usr/local/sbin/pkg install -qy py3-pip' returned 1:
pkg: No packages available to install matching 'py3-pip' have been found in the repositories
```
Here the package is called:
```
$ pkg search py|grep pip
py311-pip-23.3.2_3             Tool for installing and managing Python packages
```

tj-cisco:
Allow setting `python::version` in hiera to an integer value to match the `default` option in `install.pp`. This is required for FreeBSD.

Fixes #731
Semi related to #688 as the solution in this PR does not work when configured via hiera.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1212,5 +1212,14 @@ Alias of `Pattern[/^(<|>|<=|>=|==) [0-9]*(\.[0-9]+)*$/, /\Alatest\Z/]`
 
 Match all valid versions for python
 
-Alias of `Pattern[/\A(python)?[0-9](\.?[0-9])*/, /\Apypy\Z/, /\Asystem\Z/, /\Arh-python[0-9]{2}(?:-python)?\Z/]`
+Alias of
+
+```puppet
+Variant[Integer, Pattern[
+    /\A(python)?[0-9](\.?[0-9])*/,
+    /\Apypy\Z/,
+    /\Asystem\Z/,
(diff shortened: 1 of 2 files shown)


alexjfisher:
Was there an issue quoting the version 'number' in hiera?  Version numbers really are strings and not numbers. They just sometimes look like integers or perhaps floats, but `1.0.0` could never be either.

tj-cisco:
> Was there an issue quoting the version 'number' in hiera? Version numbers really are strings and not numbers. They just sometimes look like integers or perhaps floats, but `1.0.0` could never be either.

TBH we never tried quoting it, but it probably would have matched the first option in the pattern. This fix doesn't address `python::version: 3.9` as an unquoted string which the parser will treat as a `Float`, a semvar style version number will be treated as a string by the parser when it is unquoted which will match the first option in the pattern as the "python" prefix is optional.

PR #742, opened 2026-01-01
Add support for FreeBSD 15

smortex:


lbetz:
FacterDB still missing support for FreeBSD 15, voxpupuli/facterdb#452

PR #747, opened 2026-03-27
Add support for el10

qha:

Add support for el10.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,10 +1,9 @@
 # frozen_string_literal: true
 
-# Managed by modulesync - DO NOT EDIT
-# https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
-
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-configure_beaker(modules: :metadata)
+configure_beaker(modules: :metadata) do |host|
+  host.install_package('gnupg2') if fact('os.family') == 'RedHat' && fact('os.release.major').to_i >= 10
+end
 
 Dir['./spec/support/acceptance/**/*.rb'].sort.each { |f| require f }
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,6 +1,6 @@
 ---
 spec/spec_helper_acceptance.rb:
-  unmanaged: false
+  unmanaged: true
 .puppet-lint.rc:
   enabled_lint_checks:
     - parameter_types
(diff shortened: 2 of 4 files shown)


qha:
The ci problems look very relevant but i'm not sure what to do about them.

I suspect we need a change to spec_helper_acceptance.rb similar to what the [el10 support pr for epel added (installing gnupg2)](https://github.com/voxpupuli/puppet-epel/pull/173/changes#diff-7539b3f683671b4bc9db57fefbffb90e4b26388bac632dc639065d6e37f5b1d3) but limited to osfamily Redhat too but spec_helper_acceptance.rb looks a bit different here and most of all warns it's managed by modulesync.

I would appreciate any hints or tips!

qha:
I got some advice from a colleague so i've pushed a new variant that passes ci. If unmanaging spec_helper_acceptance.rb is not acceptable then i'll need a good deal of hints to "find another way".

spec/spec_helper_acceptance.rb
```diff
@@ -1,10 +1,9 @@
 # frozen_string_literal: true
 
-# Managed by modulesync - DO NOT EDIT
-# https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
-
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-configure_beaker(modules: :metadata)
+configure_beaker(modules: :metadata) do |host|
+  host.install_package('gnupg2') if fact('os.family') == 'RedHat' && fact('os.release.major').to_i >= 10
```
evgeni, line 6: I really wish we didn't have to do this here.

@bastelfreak didn't we want to fix that in hostgenerator?

evgeni (approved):

spec/spec_helper_acceptance.rb
evgeni: The reality is, [this line in puppet-epel](https://github.com/voxpupuli/puppet-epel/blob/cb8d7be6a1e47a76021dd181b29cbf9b630b6be8/manifests/rpm_gpg_key.pp#L17) should not use gpg on el10

bastelfreak: I don't know how to properly fix the epel module (or why it even exists instead of installing the epel-release package), but we could fix this in beaker-hostgenerator if we want:
* https://github.com/voxpupuli/beaker-hostgenerator/pull/462
* tested: https://github.com/voxpupuli/puppet-python/pull/750

bastelfreak:
closed in favour of https://github.com/voxpupuli/puppet-python/pull/750

